PR #2, opened 2017-04-27
Code generator for the REST API

7AC:
Most of the code is in the generator package. The generator reads the
JSON spec for the rest API from the rest-api-spec dir, and fits the
results into the templates stored under generator/templates.

There are 3 templates:
* method.tmpl: defines each method in the API. An example would
be the Index() method
* package.tmpl: defines the Go packages that the methods are stored in.
The packages map 1:1 with the namespaces in the API spec, rooted in the
"api" package. So the api package contains a constructor New() that
creates a *api.API object, on which you can do things like a.Index().
On the other hand the indices.create method for instance, would go in the api/indices
package. This template also defines the constructors for the API objects
(i.e., api.New(), indices.New(), etc.).
* option.tmpl: defines the optional parameters, expressed as functional
options (e.g., api.WithIndex()).

This commit also contains some basic transport code that doesn't do much
at the moment, but the client.Client object embeds the api.API object
making it easy to connect and make API calls from the same starting
point.


README.md
```diff
@@ -1 +1,50 @@
...
+    git clone https://github.com/elastic/go-elasticsearch.git
+    go install ./client
+
+## Usage
+
+```go
+import "github.com/elastic/go-elasticsearch/client"
+
+c, _ := client.New(client.WithHosts([]string{"https://elasticseach:9200"}))
+	body := map[string]interface{}{
```
ruflin: indentation seems to be off

README.md
```diff
@@ -1 +1,50 @@
...
+		"query": map[string]interface{}{
+			"term": map[string]interface{}{
+				"user": "kimchy",
+			},
+		},
+	}
+	resp, err := c.Search(body)
+```
+
+## License
```
ruflin: I suggest to put the license in a separate `LICENSES` file

api/cat/option.go
```diff
@@ -0,0 +1,585 @@
...
+)
+
+// Option is a non-required API option that gets applied to an HTTP request.
+type Option struct {
+	name  string
+	apply func(r *http.Request)
+}
+
+// WithActions - a comma-separated list of actions that should be returned. Leave empty to return all.
+func WithActions(actions []string) *Option {
```
ruflin: Is there a reason you always prefix it by `With`?

client/client.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/elastic/go-elasticsearch/transport"
+)
+
+// TODO: we should find a trick to link to api here
+
+// Client is the top-level client.
+type Client struct {
+	// API is the root of the API. Since it's embedded, it allows to call APIs directly on the client.
+	//	client, _ := client.New()
+	//	resp, err := client.Search(...)
```
ruflin: That is nice.

README.md
7AC: 👍

README.md
7AC: Yeah that's a good idea, I lifted the format from the ruby client

api/cat/option.go
7AC: Mainly to avoid collision, some of the names are both options and methods (e.g., `index`)

7AC:
Thanks for the review @ruflin, adding comments for the generator is something I definitely want to do. We can also do a live walkthrough if you're interested. For dependencies I used `dep`, but since I'm basically only adding dependencies for the generator I managed addition/cleanup manually so far (there's only 3-4).

api/bulk.go
```diff
@@ -0,0 +1,32 @@
...
+	"fmt"
+	"net/http"
+	"net/url"
+)
+
+// Bulk makes it possible to perform many index/delete operations in a single API call. See https://www.elastic.co/guide/en/elasticsearch/reference/5.x/docs-bulk.html for more info.
+//
+// body: the operation definition and data (action-data pairs), separated by newlines.
+//
+// options: optional parameters. Supports the following functional options: WithIndex, WithType, WithSource, WithSourceExclude, WithSourceInclude, WithFields, WithPipeline, WithRefresh, WithRouting, WithTimeout, WithTypeParam, WithWaitForActiveShards, WithErrorTrace, WithFilterPath, WithHuman, WithPretty, WithSourceParam, see the Option type in this package for more info.
```
andrewkroh: Is it possible to wrap long lines line this in the templates?

7AC: Unfortunately I don't think so. It's also not trivial to hack around it since for instance "body" comes from the param object (Name field) and the description comes mostly from the spec (after some formatting/enhancing). I think the easiest to handle this is to write a script that does it for you after the fact, let me look around for one I bet there is.

PR #8, opened 2017-08-16
update readme example

hothero:


tsg:
Hi @hothero, thanks for contributing here. The main contributor to this repository has left Elastic, so please expect some delays in the reviews.

Can you sign the [CLA](http://www.elasticsearch.org/contributor-agreement/), please, you can click through it and it should only take a couple of minutes.

README.md
```diff
@@ -23,13 +23,13 @@ or install it from a source code checkout:
 ```go
 import "github.com/elastic/go-elasticsearch/client"
 
-c, _ = client.New(client.WithHosts([]string{"https://elasticsearch:9200"}))
+c, _ = client.New(client.WithHost([]string{"https://elasticsearch:9200"}))
```
sweetca: several problems here:
1. `WithHost`  accept `string` as param, not array of strings `[]string`
2. variable `c` should be inited like `c := ...`
3. bad practice ignoring `error`, better to check it for `nil`

PR #9, opened 2017-09-22
Allow specifying transport, make HTTP client public so that it can be explicitly set

aeijdenberg:
This allows clients that need basic auth to provide a client that adds it.


tsg:
Hi @aeijdenberg, thanks for contributing here. The main contributor to this repository has left Elastic, so please expect some delays in the reviews.

I just merged #3 which caused a conflict in this PR. Could rebase to master, please? Also, please sign the [CLA](http://www.elasticsearch.org/contributor-agreement/), you can click through it and it should only take a couple of minutes.

PR #21, opened 2019-02-12
replace addressesFromEnvironment to urlsFromEnvironment

saromanov:
Hi! At the `NewClient` method, i replaced `addressesFromEnvironment` to `urlsFromEnvironment` for return parsed urls instead addresses. Because at the second step, its again iterating over the addresses. And trying to parse urls. In my PR, i made it at the one loop. Please review this

### DIFF
--- a/elasticsearch.go
+++ b/elasticsearch.go
@@ -51,24 +51,15 @@ func NewDefaultClient() (*Client, error) {
 // environment variable.
 //
 func NewClient(cfg Config) (*Client, error) {
-	var (
-		urls []*url.URL
-		tran estransport.Interface
-	)
+	var tran estransport.Interface
 
-	addrs := addressesFromEnvironment()
-
-	if len(addrs) > 0 && len(cfg.Addresses) > 0 {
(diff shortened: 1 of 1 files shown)


karmi:
Hi, thanks for the patch! I'm out-of-office this week, I'll have a look next week.

I think the motivation there was to keep the `addressesFromEnvironment()` very simple, and not perform the parsing. To prevent the duplication, it might be better to add another unexported helper method...

(I've removed some noisy comments from Codecov.)

karmi:
Hello, thinking about this a bit more, I still think we would mix too much concerns into the `urlsFromEnvironment()` function. I'd rather keep the `addressesFromEnvironment()` method — perhaps shortening it to `addrsFromEnvironment()` —, and extracted the conversion logic into a separate method, eg. `addrsToURLs()`.

This would remove the duplicated `range` loop, which I guess was your concern with the patch?

karmi:
Hello, a similar change has been merged as part of https://github.com/elastic/go-elasticsearch/pull/29, in 88429b3306631779e090fb4a0df757a31c47368f. Is it OK if we close this one?

karmi:
Hi @saromanov , closing this one in favour of #29, thanks for looking into it!

PR #26, opened 2019-02-27
Set TLS MinVersion not MaxVersion

andrewkroh:
Let the examples pin the minimum supported TLS version, not the maximum. This
will prevent connections to TLS 1.0.

I removed InsecureSkipVerify from the examples so that users to don't inadvertently
copy this insecure setup into their applications. Alternatively we could set the value to false
in these examples.


karmi:
I've updated the example in the `README` as well, and merged as 35433b7.

Many thanks for the patch, @andrewkroh, and sorry it took me so long to get to it...

PR #27, opened 2019-02-27
Format time duration to nanos

### LINKED ISSUE #25: Invalid time duration formatting
I was trying to do a scroll query and hit an issue caused by the formatting of time durations that are passed to Elasticsearch. You can see and example in: 

https://play.golang.org/p/fKd7ErkUTJB

I tracked down the line in the generator but I wasn't exactly sure how to regenerate the source. (I think this would be a good topic for the readme or maybe a contributing doc.)

The scroll time was being formatted based on the Go time.Duration format. So for one minute it would use 16666h40m0s. Elasticsearch supports nanoseconds as a [time unit](https://www.elastic.co/guide/en/elasticsearch/reference/6.6/common-options.html#time-units) so it can pass the Go time.Duration value through without any conversion just by appending "nanos". 

```patch
diff --git a/internal/cmd/generate/commands/gensource/generator.go b/internal/cmd/generate/commands/gensource/generator.go
index cf05218..50b7936 100755
--- a/internal/cmd/generate/commands/gensource/generator.go
+++ b/internal/cmd/generate/commands/gensource/generator.go
@@ -625,7 +625,7 @@ func (r ` + g.Endpoint.MethodWithNamespace() + `Request) Do(ctx context.Context,
                                fieldValue = `strings.Join(r.` + fieldName + `, ",")`
                        case "time.Duration":
                                fieldCondition = `r.` + fieldName + ` != 0`
-                               fieldValue = `time.Duration(r.` + fieldName + ` * time.Millisecond).String()`
+                               fieldValue = `strconv.FormatInt(int64(r.` + fieldName +`), 10) + "nanos"`
                        default: // interface{}
                                fieldCondition = `r.` + fieldName + ` != nil`
                                // TODO: Use type switching instead?
```

andrewkroh:
The scroll time was being formatted based on the Go time.Duration format. So for one minute
it would use 16666h40m0s. Elasticsearch supports nanoseconds as a time unit so so it can
pass the Go time.Duration value through without any conversion just by appending "nanos".

This updates the generator but doesn't regenerate the API sources.

Fixes #25

### DIFF
--- a/internal/cmd/generate/commands/gensource/generator.go
+++ b/internal/cmd/generate/commands/gensource/generator.go
@@ -625,7 +625,7 @@ func (r ` + g.Endpoint.MethodWithNamespace() + `Request) Do(ctx context.Context,
 				fieldValue = `strings.Join(r.` + fieldName + `, ",")`
 			case "time.Duration":
 				fieldCondition = `r.` + fieldName + ` != 0`
-				fieldValue = `time.Duration(r.` + fieldName + ` * time.Millisecond).String()`
+				fieldValue = `strconv.FormatInt(int64(r.` + fieldName +`), 10) + "nanos"`
 			default: // interface{}
 				fieldCondition = `r.` + fieldName + ` != nil`
 				// TODO: Use type switching instead?


karmi:
I finally had a proper look into this, sorry for the delay. This is a critical bug indeed, probably coming from some naïve sketching à la `time.Duration(100)`, where the value has been passed from outside.

I re-generated the API source and ran the test suite against current `7.0.0-SNAPSHOT`, to verify the tests are not impacted. It looks like the YAML test suite doesn't actually use the `timeout` values at all.

Thinking about it more broadly, I have two concerns now:

1. I think that we "leak" too much implementation here, with the `strconv.FormatInt(int64(r.Timeout), 10) + "nanos"` part. Wouldn't it be better to create a helper method in the `esapi` package, like eg. the `BoolPtr()` one, and call it? It would make the code more readable, and hide the implementation (in the good sense of the word).

2. While it's correct that we can use the `nanos` unit, as it is supported by Elasticsearch, I can see how people would be confused when they see it in a logging output, for example. After all, a timeout of one hour would be `3600000000000nanos`. Therefore, what if we would convert the duration to milliseconds, with something like this: `strconv.FormatInt(int64(1*time.Hour)/int64(time.Millisecond), 10)+"ms")`. (I don't think that, practically, anybody would set an HTTP timeout to a nanosecond resolution.)

What do you think? I can work on the implementation, but would like to hear your thoughts.

andrewkroh:
You're probably right that no one will require nanosecond precision. Regarding (1), is this a helper method to simplify the generated code? Or is this not what you mean? You could wrap the `time.Duration` in a custom type that marshals to milliseconds only. Like

```
type Duration time.Duration
func (d Duration) MarshalText() ([]byte, error) { return /* N ms */ }
```

karmi:
Correct, a helper method to simplify the generated code.

I wouldn't like to use a custom type (wrapper), because that would prevent users from simply using the stdlib's `time.Duration` one.

So I'm thinking about a function like this in the `esapi` package:

```golang
func FormatDuration(d time.Duration) string {
	if d <= time.Millisecond {
		return d.String()
	}
	return strconv.FormatInt(int64(d)/int64(time.Millisecond), 10) + "ms"
}

fmt.Println("1hour |", FormatDuration(time.Duration(time.Hour)))
// => 1hour | 3600000ms
```

Working version at https://play.golang.org/p/V_8kGdsQ0RY.

andrewkroh:
> I wouldn't like to use a custom type (wrapper), because that would prevent users from simply using the stdlib's time.Duration one.

Right, I was thinking you would keep the user-facing API as is using `time.Duration`. But when building up the `params` map you would populate it with `params["timeout"] = esapi.Duration(r.Timeout).String()`. (I just now realized that `params` is a `map[string]string` so disregard my earlier comment about `MarshalText`.)

Having your own `FormatDuration` is a fine solution too. I personally wouldn't rely on `time.Duration.String()` just to be insulated from upstream formatting changes. And make sure you account for microseconds.

I added my `Duration` type example to your example: https://play.golang.org/p/oAofKWqXx2f

karmi:
Many thanks for updating the playground code, Andrew. I've lifted the test cases and the implementation, and created a pull request at https://github.com/elastic/go-elasticsearch/pull/36.

I went with the helper method, unexported for now, `esapi.formatDuration()`. Can you have a look at that patch? I'll update the generator after it's merged.

PR #29, opened 2019-03-04
Feature: Simplify `NewClient()`

tty2:
Simplified `NewClient` function,
Added `addrsToURLs` handler,

!!! Pay attention: And I've changed package `elasticsearch_test` to `elasticsearch` in order to make tests for not importable functions.

### DIFF
--- a/elasticsearch.go
+++ b/elasticsearch.go
@@ -1,6 +1,7 @@
 package elasticsearch // import "github.com/elastic/go-elasticsearch"
 
 import (
+	"errors"
 	"fmt"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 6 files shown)


karmi:
Thanks for the patch, @tty2! I'll look at it later this week, or early next — busy with something else at the moment.

And thanks for removing the `testify` dependency, I'd like to keep the package dependency-free, unless strictly neccessary otherwise.

karmi:
I finally found some time to look at the changes. Many thanks for the patches!

I did a bit of surgery on the commits: rearranged them, squashed some changes into one, and tweaked the commit messages.

I've also tweaked the `TestAddrsToURLs` test — it more or less tests the stdlib's logic, but I think it's good to have as a placeholder, in case we discover any bugs or edge cases in future.

The gist of changes is in the linked commit 88429b3306631779e090fb4a0df757a31c47368f. Thanks again!

Issue #35, opened 2019-03-15
timeout 参数 BUG


as0x:
使用超时选项，s.client.Search(s.client.Search.WithIndex("index"), s.client.Search.WithBody(body), s.client.Search.WithTimeout(180000))
生成的请求URL为：
/index/_search?timeout=3m0s
ES 6.4 版本不能解析timeout参数，响应的错误信息为：
"reason": "failed to parse [3m0s]"
请求URL修改为 /index/_search?timeout=3m 则可以正常工作。
希望可以帮助到你们。

karmi:
The timeout formatting is currently incorrect, there's https://github.com/elastic/go-elasticsearch/pull/27 open to address.

as0x:
> The timeout formatting is currently incorrect, there's #27 open to address.

Thank you for your reply.

karmi:
This has been closed via https://github.com/elastic/go-elasticsearch/commit/c7dbbb6ef673cfbd47d24ab7acaa45e75fe2b17b and https://github.com/elastic/go-elasticsearch/commit/6b958862886db89d5dffa052185781dc54c91550 — please try with current `master` or version branches.

PR #36, opened 2019-03-16
Fix `time.Duration` conversion to Elasticsearch's time unit

karmi:
As reported in #25 and #35, the current conversion of `time.Duration` into an Elasticsearch [time unit](https://www.elastic.co/guide/en/elasticsearch/reference/current/common-options.html#time-units) is broken.

This patch builds on the implementation and discussion in #27, and adds a `esapi.formatDuration()` unexported helper function to be used in the generated code.

### DIFF
--- a/esapi/esapi_integration_test.go
+++ b/esapi/esapi_integration_test.go
@@ -1,0 +1,38 @@
+// +build integration
+
+package esapi_test
+
+import (
+	"encoding/json"
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/elastic/go-elasticsearch"
+)
+
+func TestAPI(t *testing.T) {
+	t.Run("Search", func(t *testing.T) {
+		es, err := elasticsearch.NewDefaultClient()
+		if err != nil {
+			t.Fatalf("Error creating the client: %s\n", err)
+		}
+
+		res, err := es.Search(es.Search.WithTimeout(500 * time.Millisecond))
+		if err != nil {
+			t.Fatalf("Error getting the response: %s\n", err)
+		}
+		defer res.Body.Close()
+
+		if res.IsError() {
+			t.Fatalf("Error response: %s", res.String())
+		}
+
+		var d map[string]interface{}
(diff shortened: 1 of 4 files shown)


esapi/esapi_integration_test.go
```diff
@@ -0,0 +1,39 @@
...
+		if res.IsError() {
+			t.Fatalf("Error response: %s", res.String())
+		}
+
+		var d map[string]interface{}
+		err = json.NewDecoder(res.Body).Decode(&d)
+		if err != nil {
+			log.Fatalf("Error parsing the response: %s\n", err)
+		}
+		fmt.Printf("took=%vms\n", d["took"])
```
andrewkroh: For logging in test cases I think it’s most common to use the `t.Log` and `t.Logf` functions. The logging is enabled when the `-v` flag is enabled or if the test case fails.

esapi/esapi_integration_test.go
```diff
@@ -0,0 +1,39 @@
...
+	"time"
+
+	"github.com/elastic/go-elasticsearch"
+)
+
+func TestAPI(t *testing.T) {
+	t.Run("Search", func(t *testing.T) {
+		es, err := elasticsearch.NewDefaultClient()
+		if err != nil {
+			log.Fatalf("Error creating the client: %s\n", err)
```
andrewkroh: Use `t.Fatalf` instead of `log`. I think `log.Fatal` kills the whole runtime with `os.Exit(1)` so no other tests will execute.

karmi: Right, this is a good catch, will fix!

esapi/esapi_integration_test.go
karmi: Here the `fmt` use is intentional. I use https://github.com/kyoh86/richgo as an enhanced runner, and it displays content printed by `fmt` and friends as "faint". The `t.Log/f` output is displayed prominently, so I usually not use it, since it interferes with the test error output...

Issue #42, opened 2019-04-10
Query DSL support


Youhar:
It'd be great to avoid handling huge `strings.Builder` for Search queries.
Is a functional client similar to the one in [olivere/elastic](https://github.com/olivere/elastic/wiki/QueryDSL) in the roadmap and, if so, what is the timeline for this feature?

karmi:
This package is intentionally a low-level client, while `olivere/elastic` is a high-level client with extensions for building the requests and deserializing the responses. We are aiming for offering a more high-level API in the future, but — as I've indicated in other tickets — no sooner than a machine-readable, formal specification of the request and response bodies is available.

That is especially true for the QueryDSL — in order to be able to promise a strong compatibility contract, there needs to be a formal specification to generate the Go API from.

With that in mind, though, I've been thinking about adding some kind of helper methods for similar use cases (search definition, bulk building, ...). Something like `esapi.JSONReader()`, which would automatically serialize a `map[string]interface{}` passed to it, to lighten the load on application code.

On the other hand, I was reasonably happy when using a `strings.Builder` to build search definitions, eg. in the [`xkcd` example](https://github.com/elastic/go-elasticsearch/blob/master/_examples/xkcdsearch/store.go#L192).

Youhar:
A JSON Reader would indeed streamline the process a lot, I've built a couple search queries using a `strings.Builder` and once you start handling function score queries with multiple nested filters and boosts it gets a little messy (albeit fully functional if you just test the generated string).
Please let me know if I can help in any way in implementing a `JSONReader`, I'd be happy to 😉

karmi:
Yes, completely agreed that building a complicated search definition by hand is laborious... But it's still a long way to get to a place where the code can be generated, as it is now with the API endpoints and parameters...

I'll need to wrap up couple of other things first, but I'll ping on this ticket once I'm looking into the helpers.

olivere:
@karmi How is `JSONReader` different from [`json.Marshaler`](https://golang.org/pkg/encoding/json/#Marshaler). I mean, yes, it is more generic in that it uses `io.Writer`. But is it worth it?

E.g. tools like [`github.com/tidwall/gjson`](https://github.com/tidwall/gjson) and [`github.com/tidwall/sjson`](https://github.com/tidwall/sjson) operate on standard interfaces/data structures used by `encoding/json` (`[]byte` in that case). So I think integration with tools would be easier if we'd rely on those defined in the standard library.

[...]

PR #46, opened 2019-04-25
Add the JSONReader helper

karmi:
This patch adds a `JSONReader()` helper method to facilitate passing eg. a `map[string]string` as the body, without requiring the user to manually encode it into JSON and create eg. a `strings.Reader`:

```golang
res, err = es.Search(
	es.Search.WithIndex("test"),
	es.Search.WithBody(esutil.JSONReader(map[string]string{"query": {"match": {"title": "test"}}})),
)
```

The package defines the `JSONEncoder` interface, so it's possible for the outside code to provide
an object implementing it, in order to use eg. a third-party JSON package. The default encoder
is "encoding/json".

(Related: https://github.com/elastic/go-elasticsearch/issues/42)

### DIFF
--- a/esutil/json_reader.go
+++ b/esutil/json_reader.go
@@ -1,0 +1,63 @@
+package esutil
+
+import (
+	"bytes"
+	"encoding/json"
+	"io"
+)
+
+// JSONReader is an utility function which encodes v into JSON and returns it as a reader.
+//
+func JSONReader(v interface{}) io.Reader {
+	return &jsonReader{val: v, buf: nil}
+}
+
+// JSONEncoder defines the interface for custom JSON encoders.
+//
+type JSONEncoder interface {
+	EncodeJSON(io.Writer) error
+}
+
+type jsonReader struct {
+	val interface{}
+	buf interface {
+		io.ReadWriter
+		io.WriterTo
+	}
+}
+
+func (r *jsonReader) Read(p []byte) (int, error) {
+	if err := r.initialize(); err != nil {
+		return 0, err
+	}
+	return r.buf.Read(p)
+}
+
+func (r *jsonReader) WriteTo(w io.Writer) (int64, error) {
+	if err := r.initialize(); err != nil {
+		return 0, err
+	}
+	return r.buf.WriteTo(w)
+}
+
+func (r *jsonReader) initialize() error {
+	if r.buf == nil {
+		r.buf = new(bytes.Buffer)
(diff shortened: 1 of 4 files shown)


esutil/json_reader.go
```diff
@@ -0,0 +1,48 @@
...
+//
+type JSONEncoder interface {
+	EncodeJSON(io.Writer) error
+}
+
+type jsonReader struct {
+	val interface{}
+	buf io.ReadWriter
+}
+
```
axw: You probably also want to implement https://golang.org/pkg/io/#WriterTo, which will be used by `io.Copy`. That method would avoid allocating an intermediate buffer.

karmi: Thanks for the suggestion! That makes a lot of sense, I'll have a look into that.

axw: Something like this is what I had in mind:

```diff
diff --git a/esutil/json_reader.go b/esutil/json_reader.go
index 2193515..962deab 100644
--- a/esutil/json_reader.go
+++ b/esutil/json_reader.go
@@ -34,30 +34,42 @@ func (r *jsonReader) Read(p []byte) (int, error) {
 }

 func (r *jsonReader) WriteTo(w io.Writer) (int64, error) {
-       if err := r.initialize(); err != nil {
-               return 0, err
-       }
-       return r.buf.WriteTo(w)
+       cw := countingWriter{Writer: w}
+       err := r.encode(&cw)
+       return int64(cw.n), err
 }

 func (r *jsonReader) initialize() error {
        if r.buf == nil {
                r.buf = new(bytes.Buffer)
-               return r.encode()
+               return r.encode(r.buf)
        }
        return nil
 }

-func (r *jsonReader) encode() error {
+func (r *jsonReader) encode(w io.Writer) error {
        var err error

        if e, ok := r.val.(JSONEncoder); ok {
-               err = e.EncodeJSON(r.buf)
+               err = e.EncodeJSON(w)
                if err != nil {
                        return err
                }
                return nil
        }

-       return json.NewEncoder(r.buf).Encode(r.val)
+       return json.NewEncoder(w).Encode(r.val)
+}
+
+type countingWriter struct {
+       io.Writer
+       n int
+}
+
+func (cw *countingWriter) Write(p []byte) (int, error) {
+       n, err := cw.Writer.Write(p)
+       if n > 0 {
+               cw.n += n
+       }
+       return n, err
 }
```

PR #59, opened 2019-06-18
Fix request body for esutil.JSONReader

Youhar:
See [this comment](https://github.com/elastic/go-elasticsearch/issues/42#issuecomment-503195006)


Youhar:
Already addressed through 82fc17d

PR #60, opened 2019-06-21
Fix SIGSEGV on connection refused

amenzhinsky:
Fixes SIGSEGV that occurs when `http.Do` returns an error, for example `connection refused`

### DIFF
--- a/estransport/estransport.go
+++ b/estransport/estransport.go
@@ -113,7 +113,7 @@ func (c *Client) Perform(req *http.Request) (*http.Response, error) {
 			}
 		}
 		if c.logger.ResponseBodyEnabled() {
-			if res.Body != nil && res.Body != http.NoBody {
+			if res != nil && res.Body != nil && res.Body != http.NoBody {
 				b1, b2, _ := duplicateBody(res.Body)
 				dupRes.Body = b1
 				res.Body = b2


karmi:
Hi, do you have a code example when the SIGSEGV occurs, and the line in the backtrace?

I'm asking because this shouldn't happen — eg. when you point the client to an inaccessible URL:

```
ELASTICSEARCH_URL=http://localhost:1234 go run _examples/main.go 
Error getting response: dial tcp [::1]:1234: connect: connection refused
```

Returning right away after roundtrip would make the error not accessible to the logger, see this line:
https://github.com/elastic/go-elasticsearch/blob/8ed748708d21347462d37f4a7cdd1ca267a28436/estransport/estransport.go#L122

amenzhinsky:
Got the idea, it happened only with `EnableResponseBody` flag enabled, there's no check that `res` is not nil before accessing `res.Body`. Updated the MR.

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x48 pc=0x6ebac5]

goroutine 1 [running]:
github.com/elastic/go-elasticsearch/v7/estransport.(*Client).Perform(0xc000110540, 0xc000124000, 0xc00009e2ee, 0x1, 0x0)
        /go/pkg/mod/github.com/elastic/go-elasticsearch/v7@v7.1.1/estransport/estransport.go:116 +0x535
github.com/elastic/go-elasticsearch/v7/esapi.InfoRequest.Do(0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x81f980, 0xc000110540, ...)
        /go/pkg/mod/github.com/elastic/go-elasticsearch/v7@v7.1.1/esapi/api.info.go:85 +0x3ae
```

karmi:
Thanks for the update — that's the real fix, actually. I've merged it into `master` and the version branches.

PR #61, opened 2019-06-23
Add support for custom request HTTP headers

karmi:
This patch adds support for custom HTTP headers into API calls, like this:

```golang
es.Info(
	es.Info.WithHeader(map[string]string{
		"Accept": "application/yaml"},
	),
),
```

There's been some interest in a feature like this, historically, and also, the other official clients allow to set custom HTTP headers in one way or another. I've been postponing the implementation, because setting custom headers is possible by customizing the transport, and I wasn't sure about the performance implications.

However, in order to run the [Elasticsearch REST tests](https://github.com/elastic/elasticsearch/tree/master/rest-api-spec/src/main/resources/rest-api-spec/test), setting a header when creating the request is the least complicated way, and that's a compelling reason to add the support.

From a cursory benchmark comparison, it doesn't appear like it adds weight to the existing implementation, but I would like to double-check that.

Also, the `Header` struct field is lazy initialized in the `WithHeader()` method, and while in my mental calculation it should be concurrency safe, I need to triple-check that assumption.

The interface itself follows the path of "least surprise": the `WithHeader()` takes a `map[string]string`, and uses `(http.Header).Add()` to set the field values. The field itself is of type `http.Header`, so setting it directly involves something like this:

```golang
req := esapi.InfoRequest{}
req.Header = make(http.Header)
req.Header.Add("Accept", "application/yaml")
```

The implementation doesn't deal with conflicting headers yet, eg. when the client is configured with username/password, and the request `Authorization` header is set. I think the only "least surprising" approach is to make request headers take precedence.

### DIFF
--- a/esapi/api.bulk.go
+++ b/esapi/api.bulk.go
@@ -147,7 +147,11 @@ func (r BulkRequest) Do(ctx context.Context, transport Transport) (*Response, er
 	}
 
 	if len(r.Header) > 0 {
-		req.Header = r.Header
+		for k, vv := range r.Header {
+			for _, v := range vv {
+				req.Header.Add(k, v)
+			}
+		}
 	}
 
 	if ctx != nil {
--- a/esapi/api.bulk.go
+++ b/esapi/api.bulk.go
@@ -147,9 +147,13 @@ func (r BulkRequest) Do(ctx context.Context, transport Transport) (*Response, er
 	}
 
 	if len(r.Header) > 0 {
-		for k, vv := range r.Header {
-			for _, v := range vv {
-				req.Header.Add(k, v)
+		if len(req.Header) == 0 {
+			req.Header = r.Header
+		} else {
+			for k, vv := range r.Header {
+				for _, v := range vv {
+					req.Header.Add(k, v)
+				}
 			}
 		}
 	}
--- a/esapi/api.bulk.go
+++ b/esapi/api.bulk.go
@@ -5,6 +5,7 @@ package esapi
 import (
 	"context"
 	"io"
+	"net/http"
 	"strings"
 	"time"
 )
@@ -48,6 +49,8 @@ type BulkRequest struct {
 	Human      bool
 	ErrorTrace bool
 	FilterPath []string
+
+	Header http.Header
 
 	ctx context.Context
 }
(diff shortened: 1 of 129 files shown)


estransport/estransport.go
```diff
@@ -86,9 +86,12 @@ func (c *Client) Perform(req *http.Request) (*http.Response, error) {
 	}
 
 	c.setURL(u, req)
-	c.setBasicAuth(u, req)
 	c.setUserAgent(req)
 
+	if _, ok := req.Header["Authorization"]; !ok {
```
axw: I suppose it's sensible for request headers to override headers defined based on client/transport configuration... but the auth headers may be based on env vars, which one might expect override headers defined in code? Either way this behaviour should be well documented, it'll probably surprise someone either way you go :)

esapi/api.bulk.go
```diff
@@ -143,6 +146,14 @@ func (r BulkRequest) Do(ctx context.Context, transport Transport) (*Response, er
 		req.Header[headerContentType] = headerContentTypeJSON
 	}
 
+	if len(r.Header) > 0 {
+		for k, vv := range r.Header {
```
axw: perhaps you could optimise the code path where `req.Header == nil`, and in that case set `req.Header = r.Header`?

karmi: Right, that sounds like a good optimization!

karmi: ... although, on close inspection, I'm not sure if that would work, since the field is initialized in `newRequest()`:

https://github.com/elastic/go-elasticsearch/blob/dbbd052fde51991554bb0e183718c1bcda26b4bb/esapi/esapi.request.go#L36

```golang
h := make(http.Header)
fmt.Println(h == nil)
// => false
```

axw: `len(req.Header) == 0` then. Also, that means the nil checks in the generated code are unnecessary?

karmi: Right, `len(req.Header)` is a good way to check it.

> the nil checks in the generated code are unnecessary

I think they're needed, otherwise the `r.Header.Add()` call would panic. (The `r` variable is `esapi.Request`, not `http.Request` coming from `newRequest` in the `Do()` method.)

https://github.com/elastic/go-elasticsearch/blob/c0687402022f213f62de6d78132f01a5e5242ddc/esapi/api.bulk.go#L299-L304

karmi: Implemented in dc80821.

karmi:
Closed via https://github.com/elastic/go-elasticsearch/commit/7f951bad8e941b7addda5d4e276e7530bb507f5c. Many thanks for the review and conversation, @axw!

PR #67, opened 2019-07-16
[DRAFT] Add support for retrying requests

karmi:
This patch adds a draft of support for retrying the requests automatically by the client.

### DIFF
--- a/esapi/esapi.request.go
+++ b/esapi/esapi.request.go
@@ -1,13 +1,10 @@
 package esapi
 
 import (
-	"bytes"
 	"context"
 	"io"
 	"io/ioutil"
 	"net/http"
-	"net/url"
-	"strings"
 )
 
 const (
@@ -27,30 +24,13 @@ type Request interface {
 // newRequest creates an HTTP request.
 //
 func newRequest(method, path string, body io.Reader) (*http.Request, error) {
-	r := http.Request{
-		Method:     method,
-		URL:        &url.URL{Path: path},
-		Proto:      "HTTP/1.1",
-		ProtoMajor: 1,
-		ProtoMinor: 1,
-		Header:     make(http.Header),
-	}
+	req, err := http.NewRequest(method, path, body)
 
-	if body != nil {
-		switch b := body.(type) {
-		case *bytes.Buffer:
-			r.Body = ioutil.NopCloser(body)
-			r.ContentLength = int64(b.Len())
-		case *bytes.Reader:
-			r.Body = ioutil.NopCloser(body)
-			r.ContentLength = int64(b.Len())
-		case *strings.Reader:
-			r.Body = ioutil.NopCloser(body)
-			r.ContentLength = int64(b.Len())
-		default:
-			r.Body = ioutil.NopCloser(body)
+	if body != nil && req.GetBody == nil {
+		req.GetBody = func() (io.ReadCloser, error) {
+			return ioutil.NopCloser(body), nil
 		}
 	}
 
-	return &r, nil
+	return req, err
 }
--- a/esapi/esapi.request.go
+++ b/esapi/esapi.request.go
@@ -26,6 +26,10 @@ type Request interface {
 func newRequest(method, path string, body io.Reader) (*http.Request, error) {
 	req, err := http.NewRequest(method, path, body)
 
+	if err != nil {
+		return nil, err
+	}
+
 	if body != nil && req.GetBody == nil {
 		req.GetBody = func() (io.ReadCloser, error) {
 			return ioutil.NopCloser(body), nil
--- a/esapi/esapi.request.go
+++ b/esapi/esapi.request.go
@@ -1,10 +1,13 @@
 package esapi
 
 import (
+	"bytes"
 	"context"
 	"io"
 	"io/ioutil"
 	"net/http"
+	"net/url"
+	"strings"
 )
 
 const (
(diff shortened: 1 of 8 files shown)


esapi/esapi.request.go
```diff
@@ -27,30 +24,13 @@ type Request interface {
...
-			r.Body = ioutil.NopCloser(body)
-			r.ContentLength = int64(b.Len())
-		case *strings.Reader:
-			r.Body = ioutil.NopCloser(body)
-			r.ContentLength = int64(b.Len())
-		default:
-			r.Body = ioutil.NopCloser(body)
+	if body != nil && req.GetBody == nil {
+		req.GetBody = func() (io.ReadCloser, error) {
+			return ioutil.NopCloser(body), nil
```
axw: This won't work, you're just returning the same `body` over and over. The first attempt will consume it, and then it'll be empty on retries. Instead you'd need to modify `req.Body` so that it reads into a buffer, and then have GetBody use that buffer. Basically, move the `dupReqBodyForLog` logic here.

esapi/esapi.request.go
```diff
@@ -27,30 +24,13 @@ type Request interface {
...
-			r.ContentLength = int64(b.Len())
-		case *bytes.Reader:
-			r.Body = ioutil.NopCloser(body)
-			r.ContentLength = int64(b.Len())
-		case *strings.Reader:
-			r.Body = ioutil.NopCloser(body)
-			r.ContentLength = int64(b.Len())
-		default:
-			r.Body = ioutil.NopCloser(body)
+	if body != nil && req.GetBody == nil {
```
axw: if `err != nil`, this will panic. You should just do the usual `if err != nil { return nil, err }` here.

estransport/estransport.go
```diff
@@ -84,41 +85,90 @@ func New(cfg Config) *Client {
 //
 func (c *Client) Perform(req *http.Request) (*http.Response, error) {
 	var (
-		dupReqBody io.Reader
+		res *http.Response
+		err error
+
+		dupReqBodyForLog io.ReadCloser
```
axw: the idea behind using `GetBody` was to get rid of this

estransport/estransport.go
```diff
@@ -84,41 +85,90 @@ func New(cfg Config) *Client {
...
-	c.setURL(u, req)
 	c.setUserAgent(req)
-	c.setAuthorization(u, req)
 
-	// Duplicate request body for logger
-	//
-	if c.logger != nil && c.logger.RequestBodyEnabled() {
-		if req.Body != nil && req.Body != http.NoBody {
-			dupReqBody, req.Body, _ = duplicateBody(req.Body)
+	for i := 1; ; i++ {
```
axw: `if i := 0; i < maxRetries; i++ {` ?

estransport/estransport.go
```diff
@@ -84,41 +85,90 @@ func New(cfg Config) *Client {
...
+
+		// Break if retries have been exhausted
+		//
+		if i >= maxRetries {
+			// fmt.Printf("Perform: aborting after %d attempts\n", i)
+			break
+		}
+
+		// TODO(karmi): If c.DisableRetryOnError => break
+		// TODO(karmi): Retry on status [502, 503, 504]
```
axw: here you would do something like:

```go
if req.Body != nil && req.Body != http.NoBody {
    body, err := req.GetBody()
    if err != nil {
        return nil, err
    }
    req.Body = body
}
```

estransport/estransport.go
```diff
@@ -84,41 +85,90 @@ func New(cfg Config) *Client {
...
+		// Update request
+		//
+		c.setURL(nodeURL, req)
+		c.setAuthorization(nodeURL, req)
+
+		// Duplicate request body for logger
+		//
+		if c.logger != nil && c.logger.RequestBodyEnabled() {
+			if req.Body != nil && req.Body != http.NoBody {
+				dupReqBodyForLog, req.Body, _ = duplicateBody(req.Body)
```
axw: You can use `req.GetBody` here:

```go
if req.Body != nil && req.Body != http.NoBody {
    body, err := req.GetBody()
    if err != nil {
        return nil, err
    }
    dupReqBodyForLog = body
}
```

Or you could just let `logRoundTrip` do that.

esapi/esapi.request.go
karmi: Fixed in c78709c.

karmi:
Executable example:

```golang
//+build ignore

package main

import (
	"context"
	"fmt"
	"log"
	"math"
	"net"
	"net/http"
	"os"
	"os/signal"
	"strings"
	"time"

	"github.com/elastic/go-elasticsearch/v8/esapi"

	"github.com/cenkalti/backoff"
)

var (
	_ = fmt.Print
	_ = context.WithTimeout
	_ = math.Exp
	_ = strings.NewReader
	_ = http.DefaultClient
)

func main() {
	log.SetFlags(0)

	ticker := time.NewTicker(time.Second)
	defer ticker.Stop()

	aborted := make(chan os.Signal)
	signal.Notify(aborted, os.Interrupt)

	retryBackoff := backoff.NewExponentialBackOff()
	retryBackoff.InitialInterval = time.Second
	_ = retryBackoff

	cfg := elasticsearch.Config{
		Addresses: []string{
			"http://localhost:9200",
			// "http://localhost:1000",
			// "http://localhost:2000",
			// "http://localhost:3000",
			"http://localhost:4000",
			"http://localhost:5000",
			// "http://localhost:6000",
			// "http://localhost:7000",
			// "http://localhost:8000",
			// "http://localhost:9200",
			// "http://localhost:9000",
			// "http://localhost:9201",
			// "http://localhost:9202",
		},

		Logger: &estransport.ColorLogger{
			Output:             os.Stdout,
			EnableRequestBody:  true,
			EnableResponseBody: true,
		},

		// RetryOnStatus:        []int{404},
		// EnableRetryOnTimeout: true,
		// MaxRetries: 10,

		// RetryBackoff: func(i int) time.Duration {
		// 	d := time.Duration(i) * time.Second
		// 	fmt.Printf("Attempt: %d | Sleeping for %s...\n", i, d)
		// 	return d
		// },

		// RetryBackoff: func(i int) time.Duration {
		// 	d := time.Duration(math.Exp2(float64(i))) * time.Second
		// 	fmt.Printf("Attempt: %d | Sleeping for %s...\n", i, d)
		// 	return d
		// },

		RetryBackoff: func(i int) time.Duration {
			if i == 1 {
				retryBackoff.Reset()
			}
			d := retryBackoff.NextBackOff()
			fmt.Printf("Attempt: %d | Sleeping for %s...\n", i, d)
			return d
		},

		// Transport: &http.Transport{
		// 	ResponseHeaderTimeout: time.Millisecond,
		// },
	}

	es, err := elasticsearch.NewClient(cfg)
	if err != nil {
		log.Fatalf("Error creating the client: %s", err)
	}
	log.Println("Client ready with URLs:", es.Transport.(*estransport.Client).URLs())

	for {
		select {
		case <-aborted:
			log.Println("\nDone!")
			return
		case <-ticker.C:
			var (
				res *esapi.Response
				err error
			)
			// res, err = es.Info()

			// ctx, cancel := context.WithTimeout(context.Background(), time.Nanosecond)
			// defer cancel()
			// res, err = es.Info(es.Info.WithContext(ctx))

			// es.Search(es.Search.WithTimeout(time.Nanosecond))

			// res, err = es.Get("test", "MISSING")

			res, err = es.Index("test", strings.NewReader(`{"foo":"bar"}`))

			if err != nil {
				if e, ok := err.(net.Error); ok {
					log.Fatalf("Error getting response: [%T]: %s (timeout:%v)", e, e, e.Timeout())
				} else {
					log.Fatalf("Error getting response: [%T]: %s", err, err)
				}
			}
			if res.IsError() {
				log.Fatalf("Error response: %s", res.Status())
			}
		}
	}
}
```

PR #73, opened 2019-07-25
WIP: support marshaling error responses into go error for error responses from ES

zaquestion:
Wanted to get this up for initial review on implementation and if see if
this change would be considered in the first place. Also want get some
more general guidance contributing to this repo. Such as:
- Proper way to put up changes for the different ES versions this package provides
- How to test with these generated functions?

### DIFF
--- a/internal/cmd/generate/commands/gensource/generator.go
+++ b/internal/cmd/generate/commands/gensource/generator.go
@@ -749,6 +749,23 @@ func (r ` + g.Endpoint.MethodWithNamespace() + `Request) Do(ctx context.Context,
 		return nil, err
 	}` + "\n\n")
 
+	g.w(`if res.StatusCode != http.StatusOK {
+		var errResp struct {
+			Errs []struct {
+				Type    string ` + "`" + `json:"type"` + "`" + `
+				Reason  string ` + "`" + `json:"reason"` + "`" + `
+				Phase   string ` + "`" + `json:"phase"` + "`" + `
+				Grouped bool   ` + "`" + `json:"grouped"` + "`" + `
+			} ` + "`" + `json:"error"` + "`" + `
+			Status int ` + "`" + `json:"status"` + "`" + `
+		}
+		err := json.NewDecoder(res.Body).Decode(&errResp)
+		if err != nil {
+			return nil, err
+		}
+		return nil, fmt.Errorf("%d %v", errResp.Status, errResp.Errs)
+	}` + "\n\n")
+
 	// Generate the return value
 	g.w(`
 	response := Response{
--- a/esapi/api.search.go
+++ b/esapi/api.search.go
@@ -4,6 +4,7 @@ package esapi
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
 	"net/http"


internal/cmd/generate/commands/gensource/generator.go
```diff
@@ -749,6 +749,23 @@ func (r ` + g.Endpoint.MethodWithNamespace() + `Request) Do(ctx context.Context,
 		return nil, err
 	}` + "\n\n")
 
+	g.w(`if res.StatusCode != http.StatusOK {
```
zaquestion: Initial note will need to update this to follow the "non 2xx" logic in `IsError`. 

Ref https://github.com/elastic/go-elasticsearch/blob/master/esapi/esapi.response.go#L74-L76

Likewise will need to figure out what to do with `IsError` since its incompatible with these changes, I suspect it could be removed entirely, but not sure how wide spread its usage may be.

karmi:
Hey, the idea itself makes a lot of sense, but I'm afraid it has to be implemented rather differently.

First of all, the contract of the client is that "we never parse the body" — and here that contract would be broken. I think I've had something similar in the preliminary sketches of the client, but ultimately removed it.

Second, this subtly changes the semantics of the `res, err` tuple being returned: suddenly, a perfectly fine HTTP response is converted to a Go error. The idea behind the current return values is that `err` indicates a true _error_ on the transport layer (connection refused, no connection, retries exhausted, ...), and when a response is, in fact, returned from Elasticsearch, this is passed on to the calling code to deal with it. Of course, it's a matter of perspective whether eg. a `500` response constitutes an "error" (an "exception" in another language) or no, but I still think this semantical difference is the best practical solution.

Third, there's definitely a valid argument for a structured "error response" being returned. However, I see it more like the [`(*esapi.Response).String()`](https://github.com/elastic/go-elasticsearch/blob/c49377020e092848a3af5737ab122df131199632/esapi/esapi.response.go#L25) method — a kind of helper, which would parse the response body into a Go struct, for a more convenient handling in the calling code. I'm not entirely sure on the naming for that, though.

karmi:
Hi @zaquestion , should I close the PR?

zaquestion:
@karmi Hey, sorry haven't gotten back, if we could leave it open I've got a couple ideas for alternate implementation after chatting with some go friends at elastic. Will describe here and if we can settle on something will modify in this PR.

While I still believe client APIs should return any error related to the API and not just transport errors. Particularly since go favors all handling through its error values and choicefully does not implement exceptions. That said...

We could instead support:
```
func (r *Response) Error() error {
    if !r.IsError() {
        return nil
    }
    ... comparable marshaling to the current implementation in this PR ...
}
```
This way the API calls will only return transport errors and API errors can be easily checked and interacted with as errors, which would also obviate the need for excessive type assertion to get these values.

Not sure on any other alternatives at the moment, but open to suggestions if we can find something that works.

karmi:
Hey, sure, we can keep the PR open. I'll be off the grid in August, so I might not see the updates soon.

> While I still believe client APIs should return any error related to the API and not just transport errors. (...)

I'm afraid I don't know what that means, specifically. Maybe you can refine it a little bit?

Also, as I said, changing the semantics to something like _"`res` is a response, if it's eg. a `200 OK`, otherwise it's nil and the actual response is in `err`"_ is not something I would accept lightly — I consider the current distinction between `res` and `err` to work great, in my own applications and examples. I would say it respects the general Go semantics for error handling rather well.

Take this example: https://github.com/elastic/go-elasticsearch/blob/ea052088db25bb87db9ce22b8191a5205665ee99/_examples/main.go#L81-L99

Perhaps when you play with a concrete, nontrivial example like this one, it will help to guide the implementation. Notice how that example — as many others in the repository — handles separately the case of _"error getting a response"_ versus _"got an error response"_.

With all that said, I'll say again that something like `(*esapi.Response).ParsedError()` (not sure about the naming yet) is definitely on my TODO list — error response structures are rather stable in Elasticsearch, so it would be possible to provide a convenient accessors for them, compared to other responses.

PR #81, opened 2019-08-22
Fix rule to parse for CloudID.

gjbae1212:
Hi maintainer.
I have used this library with Elastic Cloud(ElasticSearch) recently.

And when I was connecting to Elastic Cloud(ElasticSearch) having GCP  backend in the Tokyo region, I used CloudID instead of an endpoint.

But I encountered an error having "invalid port".
First encountered  I don't understand this error, So I did look under the hood for troubleshooting this error.

And I found this error reason, and It is that CloudID made by GCP in the Tokyo region regardless of other backend had an odd pattern of a key.

A normal CloudID seems to a pattern like "cluster_name:base64 string", "base64 string" seems to a pattern like "host$number$number" after decode.
But "base64 string"(part of CloudID) made by GCP in the Tokyo region has a pattern "host:9243$number$number".

So I contacted the Elastic Cloud Support Center for resolving this issue.
After some days,  I was answered this issue resolve.

But I would say this library needs to defensive code which shouldn't occur an error for this issue.
Both Filebeat and Logstash have worked very well, even though this issue. 
I would say both Filebeat and Logstash seem to have a defensive code for this issue.

So I Suggest to this PR.

Thank you.

### DIFF
--- a/elasticsearch.go
+++ b/elasticsearch.go
@@ -161,7 +161,7 @@ func addrsToURLs(addrs []string) ([]*url.URL, error) {
 //
 func addrFromCloudID(input string) (string, error) {
 	var (
-		port   = 9243
+		port   = "9243"
 		scheme = "https://"
 	)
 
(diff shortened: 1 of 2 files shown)


karmi:
Hello, sorry for taking so long to respond! I've looked into the issue and consulted with Cloud, and the culprit here is adding the `9243` port by default into the parsed address. I've tested against a new deployment, and the value no longer contains port, and in fact, that's what clients should do when connecting to Cloud — leave the value empty, and let the HTTP library pick up the port, `443` in this case.

I've fixed the problem in a different way — instead of parsing the port out of the value, I just removed it (0b113bc5). I've added a couple of tests to cover the behaviour. Thanks for raising the issue and for the patch! Please re-open the issue or open a new one if you would run into problems again.

PR #84, opened 2019-08-26
Add token support for API calls

geoffmore:
The context behind this PR is my desire to support token based authentication for API calls.

The following changes were made to support token authentication according to [token documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-get-token.html).

elasticsearch.go:
- Add Token field to elasticsearch.Config struct
- In function elasticsearch.NewClient, add Token field in call to estransport.New for initialization of variable tp

estransport/estransport.go
- Add Token field to estransport.Config struct
- Add token field to estransport.Client struct
- In function estransport.New, add token field to Client return
- In `func (c *Client) setAuthorization(u *url.URL, req *http.Request) *http.Request`, add logic supporting token headers similar what is done for apikey.

### DIFF
--- a/elasticsearch.go
+++ b/elasticsearch.go
@@ -30,7 +30,9 @@ type Config struct {
 	Password  string   // Password for HTTP Basic Authentication.
 
 	CloudID string // Endpoint for the Elastic Service (https://elastic.co/cloud).
-	APIKey  string // Base64-encoded token for authorization; if set, overrides username and password.
+
+	APIKey string // Base64-encoded token for authorization; if set, overrides username, password, and token
+	Token  string // Token used for authorization. if set, overrides username and password.
 
 	Transport http.RoundTripper  // The HTTP transport object.
 	Logger    estransport.Logger // The logger object.
@@ -110,7 +112,9 @@ func NewClient(cfg Config) (*Client, error) {
 		URLs:     urls,
 		Username: cfg.Username,
 		Password: cfg.Password,
-		APIKey:   cfg.APIKey,
+
+		APIKey: cfg.APIKey,
+		Token:  cfg.Token,
 
 		Transport: cfg.Transport,
 		Logger:    cfg.Logger,
--- a/estransport/estransport.go
+++ b/estransport/estransport.go
@@ -40,7 +40,9 @@ type Config struct {
 	URLs     []*url.URL
 	Username string
 	Password string
-	APIKey   string
+
+	APIKey string
+	Token  string
 
 	Transport http.RoundTripper
 	Logger    Logger
@@ -52,7 +54,9 @@ type Client struct {
 	urls     []*url.URL
 	username string
 	password string
-	apikey   string
+
+	apikey string
+	token  string
 
 	transport http.RoundTripper
 	selector  Selector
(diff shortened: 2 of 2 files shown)


karmi:
Hello, thanks for the patch! I'm on vacation currently, I'll have a look when I'm back. The changes themselves look good, I'll just have to think through the order of precedence between the various authentication values.

geoffmore:
@karmi Did the precedence match what you expected?

karmi:
Hi, sorry for such a long delay! I've been thinking about the feature a little bit more, and I've realized it would actually make the client diverge from the other official clients, which don't support such a feature out of the box.

Since Elasticsearch itself doesn't support a token-based authentication, and it's applicable only when some kind of proxy sits in front of it, I think it qualifies as a "custom environment". Therefore, I'd suggest creating a custom transport, wrapping the default one, and setting the request header — as in the following example:

```golang
//+build ignore

package main

import (
	"fmt"
	"log"
	"net/http"
	"os"

)

const (
	serverToken = "abc123"
	clientToken = "xyz456" // <<< Change to the same value for successful authorization
)

type AuthorizedTransport struct{}

func (t *AuthorizedTransport) RoundTrip(req *http.Request) (*http.Response, error) {
	req.Header.Set("Authorization", "Bearer "+clientToken)
	fmt.Println(">>> ", req.Header.Get("Authorization"))
	return http.DefaultTransport.RoundTrip(req)
}

func main() {
	log.SetFlags(0)

	defaultHandler := func(w http.ResponseWriter, r *http.Request) {
		fmt.Println("<<< ", r.Header.Get("Authorization"))
		if r.Header.Get("Authorization") != "Bearer "+serverToken {
			http.Error(w, "Invalid credentials", 403)
			return
		}
		fmt.Fprintln(w, "OK")
	}

	go func() {
		log.Fatalln(http.ListenAndServe("localhost:10100", http.HandlerFunc(defaultHandler)))
	}()

	es, _ := elasticsearch.NewClient(
		elasticsearch.Config{
			Addresses: []string{"http://localhost:10100"},
			Transport: &AuthorizedTransport{},
			// >>>>> Wrap the transport >>>>>>>>>>>>>>>>>>>>>>>>
			Logger: &estransport.ColorLogger{Output: os.Stdout},
			// >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
		})

	res, err := es.Info()
	if err != nil {
		log.Fatalf("Error getting response: %v", err)
	}
	defer res.Body.Close()

	if res.IsError() {
		log.Fatalf("Error response: %s", res)
	}

	log.Println(res)
}
```

With this reasoning, I'd like to close the PR without merging. Please re-open or comment if you would like to discuss it further.

karmi:
Hi @geoffmore , I was wrong about Elasticsearch itself not supporting the bearer/token authentication — sorry about that.

There's a broader discussion happening in https://github.com/elastic/go-elasticsearch/issues/106#issuecomment-557060934 now about custom HTTP headers, which is related to the bearer/token authentication feature.

geoffmore:
@karmi I don't know enough about custom HTTP headers to contribute anything meaningful to #106 , but I am following the conversation to see where I can contribute.

karmi:
Custom global HTTP request headers have been added in https://github.com/elastic/go-elasticsearch/commit/3f39d542adc013c69c7818092e6ac0d428e1b9a7.

PR #95, opened 2019-10-19
Add connection pool

karmi:
This patch adds support for the "connection pool" feature, standard in official Elasticsearch clients, which maintains state for node connections passed in configuration, and removes connections with failures from the list of live connections. It periodically adds them back to the list, with an exponentional backoff delay.

The main parts of the implementation are contained in the `connection.go` file. The implementation short-circuits the connection pooling when only a single URL has been passed in configuration — common when using Elastic Cloud or a proxy/load balancer in front of Elasticsearch. When multiple URLs are passed, they are added to the list of live connections, and returned in a round-robin fashion from the `Next()` method. When the connection fails to respond to a request, it is removed from the list of live connections, put into the list of dead connections, and scheduled for a "resurrect" attempt. In case no live connection is available, a dead connection is forcibly resurrected. A `sync.Mutex` is used for synchronization.

See reference implementation in Ruby ([1](https://github.com/elastic/elasticsearch-ruby/blob/master/elasticsearch-transport/lib/elasticsearch/transport/transport/connections/collection.rb), [2](https://github.com/elastic/elasticsearch-ruby/blob/master/elasticsearch-transport/lib/elasticsearch/transport/transport/connections/connection.rb)) and Python ([1](https://github.com/elastic/elasticsearch-py/blob/master/elasticsearch/connection_pool.py#L180)).

Apart from regular unit tests, an end-to-end integration test is contained in `connection_integration_test.go`, using a "fake" server provided by the `net/http` package.

Benchmarks are provided in the `connection_benchmark_test.go` file. Example output:

```
Next()/Single____-4         	 21851822	        51.6 ns/op	       0 B/op	       0 allocs/op
Next()/Parallel_(100)-4     	  6760563	         210 ns/op	       0 B/op	       0 allocs/op
Next()/Parallel_(1000)-4    	  1689972	         860 ns/op	       0 B/op	       0 allocs/op
Remove()/Single____-4       	 72488359	        16.2 ns/op	       0 B/op	       0 allocs/op
Remove()/Parallel_(10)-4    	143871522	        8.56 ns/op	       0 B/op	       0 allocs/op
Remove()/Parallel_(100)-4   	141271904	        8.63 ns/op	       0 B/op	       0 allocs/op
Resurrect()/Single_______-4 	 17904594	        66.2 ns/op	       0 B/op	       0 allocs/op
Resurrect()/Parallel_(10)-4 	  4877097	         240 ns/op	       0 B/op	       0 allocs/op
Resurrect()/Parallel_(100)-4	  5453721	         225 ns/op	       0 B/op	       0 allocs/op
```

For easier debugging, the implementation allows to use the `EnableDebugLogger` option to print debugging statements related to connection management to STDOUT.

This patch also adds support for collecting and getting client metrics: number of requests and failures, statistics for specific response codes, and the contents of the live/dead connection lists. An example for usage with the `expvar` package is provided in `_examples/instrumentation/expvar.go`. Integration with APM and other instrumentation tools will be researched and documented in future.


estransport/connection.go
```diff
@@ -0,0 +1,300 @@
...
+		c1 := cp.dead[i]
+		c2 := cp.dead[j]
+		c1.Lock()
+		if c1 != c2 {
+			c2.Lock()
+		}
+		defer c1.Unlock()
+		if c1 != c2 {
+			defer c2.Unlock()
+		}
```
axw:
```go
c1.Lock()
defer c1.Unlock()
if c1 != c2 {
	c2.Lock()
	defer c2.Unlock()
}
```
? (Though I still stand by my statement on Slack that this is unnecessary, and `i != j` is invariant.)

estransport/connection.go
```diff
@@ -0,0 +1,300 @@
...
+
+// scheduleResurrect schedules the connection to be resurrected.
+//
+func (cp *roundRobinConnectionPool) scheduleResurrect(c *Connection) {
+	factor := func(a, b int) float64 {
+		if a > b {
+			return float64(b)
+		}
+		return float64(a)
+	}(c.Failures-1, defaultResurrectTimeoutFactorCutoff)
```
axw: `factor := math.Min(float64(c.Failures-1), defaultResurrectTimeoutFactorCutoff)` ?

karmi: Argh, embarrasing :) Fixed in f7cbe9a.

estransport/connection.go
```diff
@@ -0,0 +1,300 @@
...
+			cp.debugLogger.Logf("Already removed %s\n", c.URL)
+		}
+		c.Unlock()
+		return nil
+	}
+
+	if cp.debugLogger != nil {
+		cp.debugLogger.Logf("Removing %s...\n", c.URL)
+	}
+	c.markAsDead()
```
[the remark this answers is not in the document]
karmi: Yes, the only reason was worrying about performance. I've added locking of the pool in 5aedfbb.

estransport/connection.go
karmi: I was getting errors when debugging the locking with https://github.com/sasha-s/go-deadlock, and this seemingly unnecessary condition made them go away. I do agree that it's counterintuitive. I'll remove those, and if there's some hidden race condition / deadlock, we'll fix it along the way.

karmi: Removed in 9be9b94.

estransport/connection.go
```diff
@@ -0,0 +1,300 @@
...
+	Remove(*Connection) error
+}
+
+// Connection represents a connection to a node.
+//
+type Connection struct {
+	sync.Mutex
+
+	URL       *url.URL  `json:"url"`
+	Dead      bool      `json:"dead"`
```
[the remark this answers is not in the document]
karmi: Not sure about that, to be honest. I can't see a way for outside code to access an instance of `Connection` — they are hidden behind unexported fields of `estransport.Client`. I like that these fields (and in future, other fields, coming from cluster state) are exported, for documentation purposes, and also, when the package allows replacing connection pool implementation, which is planned. I've removed the JSON tags in c7cf16b, though.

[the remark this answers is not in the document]
axw: As soon as it's exported, it's part of the API. Maybe unexport `Connection` (rename to `connection`) until needed, and at that point it can be reassessed? If nothing else, documentation about concurrency expectations would be good.

estransport/connection.go
```diff
@@ -0,0 +1,300 @@
...
+			return c, nil
+		}
+		return nil, errors.New("no connection available")
+	}
+
+	if cp.curr >= len(cp.live) {
+		cp.curr = len(cp.live) - 1
+	}
+
+	if cp.curr < 0 {
```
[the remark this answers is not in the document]
karmi: Thanks, this is a much better code flow. I've refactored it along these lines in f35372b, apart from calling `Resurrect()`, that's something I'll need more time to investigate.

estransport/connection.go
```diff
@@ -0,0 +1,280 @@
...
+
+func newSingleConnectionPool(u *url.URL) *singleConnectionPool {
+	cp := singleConnectionPool{connection: &Connection{URL: u}}
+
+	return &cp
+}
+
+func newRoundRobinConnectionPool(u ...*url.URL) *roundRobinConnectionPool {
+	var conns []*Connection
+	for _, url := range u {
```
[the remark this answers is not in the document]
karmi: That's correct. I'm wondering, though, if it pays off — after all, connections might/will be removed and added to the list, eg. when a connection dies, or new nodes are discovered?

estransport/connection.go
```diff
@@ -0,0 +1,280 @@
...
+		conns = append(conns, &Connection{URL: url})
+	}
+
+	cp := roundRobinConnectionPool{live: conns, orig: u, curr: -1}
+
+	if cp.enableMetrics {
```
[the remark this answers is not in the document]
karmi: That's a good catch, and fair point, thanks! The thing is, I want to keep the connection pool as separate as possible, and the other thing is, adding parameters to the constructor function feels like a slippery slope. Granted, it could be rewritten to take a config struct, but that overcomplicates the calling a bit... Hard choices... I'll give it some thoughts.

estransport/connection.go
```diff
@@ -0,0 +1,280 @@
...
+			cp.debugLogger.Logf("Already resurrected %s\n", c.URL)
+		}
+		return nil
+	}
+
+	if cp.debugLogger != nil {
+		cp.debugLogger.Logf("Resurrecting %s\n", c.URL)
+	}
+
+	c.markAsLive()
```
simitt: I think it is worth to optimize for the good path. Is it feasible to do a healthcheck here on the resurrected connection, and if successful call `markAsHealthy`? This way you could get rid of any locking in the _good_ case of using a connection https://github.com/elastic/go-elasticsearch/pull/95/files#diff-8fce2da81ff9e3289d4a5515a242448bR263-R269.

karmi: With a "healthcheck" you mean something like calling `HEAD /` on the node, ie. a "ping"? That's definitely something I was thinking about, but that would need a pluggable "strategy" for the resurrection, as outlined in the TODO comment, because I don't want to hardcode the "ping" into the implementation.

simitt: That makes sense. Totally fair to do this at a later point if necessary.

estransport/connection.go
```diff
@@ -0,0 +1,280 @@
...
+	defaultResurrectTimeoutInitial      = 60 * time.Second
+	defaultResurrectTimeoutFactorCutoff = 5
+)
+
+// ConnectionPool defines the interface for the connection pool.
+//
+type ConnectionPool interface {
+	Next() (*Connection, error)
+	Remove(*Connection) error
+}
```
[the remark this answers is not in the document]
urso: What is nice about interface types is, that you don't need to export the interface in order to require it to be implemented. This means, even if you want people to provide custom implementation, there is no need to export the interface.

One question to ask is: who will be the user/consumer of a ConnectionPool implementation.

Some common pattern:

```
type myInterfaceType.

func (ifc myInterfaceType) <concrete Type> {
   ...
}
```

For reference:
- https://rakyll.org/interface-pollution/
- https://twitter.com/davecheney/status/1030790804011245569?lang=de

All in all no strong feelings here. It depends case by case basis if it makes sense to export a symbol or not (but too many symbols might lead to pollution).

estransport/connection.go
```diff
@@ -0,0 +1,280 @@
...
+	curr int           // Index of the current connection
+	live []*Connection // List of live connections
+	dead []*Connection // List of dead connections
+	orig []*url.URL    // List of original URLs, passed in during initialization
+
+	metrics       *metrics
+	enableMetrics bool
+
+	debugLogger DebuggingLogger
+}
```
urso: maybe consider moving the two implementations into separate go files.

estransport/connection.go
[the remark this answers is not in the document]
urso: Alternative choice (and not too uncommon) is to drop the constructor but require only the struct to be filled out. In this case the 'zero-value' of RoundRobinConnectionPool would be an valid instance.

I don't see anything in the struct definition which would require a constructor per se. It could become:

```

type roundRobinConnectionPool struct {
	sync.Mutex

	curr int           // Index of the current connection
	live []*Connection // List of live connections
	dead []*Connection // List of dead connections
	URLs []*url.URL    // List of original URLs, passed in during initialization

	Metrics <some interface type>
	enableMetrics bool // disable if metrics is nil?

	DebugLogger DebuggingLogger
}
```

estransport/connection.go
```diff
@@ -0,0 +1,280 @@
...
+var (
+	defaultResurrectTimeoutInitial      = 60 * time.Second
+	defaultResurrectTimeoutFactorCutoff = 5
+)
+
+// ConnectionPool defines the interface for the connection pool.
+//
+type ConnectionPool interface {
+	Next() (*Connection, error)
+	Remove(*Connection) error
```
urso: what is the purpose of Remove? Is a connection to be removed forever?

Is removed only after failure? If so, how about naming it failed?

Utilities like ConnectionPools are sensitive to the protocol. For example: do we need to full consumer if a connection failed? Do we have to close the underlying TCP connection if the connection failed? Not properly documenting the protocol and required behavior. Mixing wrong implementations can easily lead to security issues. Even better if there is a common connection pool protocol test suite.

A good read of potential issues, and how the custom connection pool in the redis golang library did lead to a security breah: https://medium.com/@cep21/interface-contracts-matter-812dd43709b3

estransport/connection.go
```diff
@@ -0,0 +1,280 @@
...
+//
+func (cp *singleConnectionPool) Next() (*Connection, error) {
+	return cp.connection, nil
+}
+
+// Remove is a no-op for single connection pool.
+//
+func (cp *singleConnectionPool) Remove(c *Connection) error {
+	return nil
+}
```
urso: would it make sense to have a cirtcuite breaker or some backoff that disables the connection for N seconds if it has been 'removed' ?

estransport/connection.go
```diff
@@ -0,0 +1,280 @@
...
+
+	// Push item to dead list and sort slice by number of failures
+	cp.dead = append(cp.dead, c)
+	sort.Slice(cp.dead, func(i, j int) bool {
+		c1 := cp.dead[i]
+		c2 := cp.dead[j]
+		c1.Lock()
+		c2.Lock()
+		defer c1.Unlock()
+		defer c2.Unlock()
```
urso: cp.dead can not be touched because of `cp.Lock`. Do we need to lock the mutexes here? Are connections in the pool supposed to be used by many go-routines concurrently?

estransport/connection.go
```diff
@@ -0,0 +1,280 @@
...
+	if cp.debugLogger != nil {
+		cp.debugLogger.Logf("Resurrect %s (failures=%d, factor=%1.1f, timeout=%s) in %s\n", c.URL, c.Failures, factor, timeout, c.DeadSince.Add(timeout).Sub(time.Now().UTC()).Truncate(time.Second))
+	}
+
+	time.AfterFunc(timeout, func() { cp.Resurrect(c) })
+}
+
+// markAsDead marks the connection as dead.
+//
+func (c *Connection) markAsDead() {
```
urso: We might consider to export these methods given that users should be allowed to implement custom pools.

estransport/estransport.go
```diff
@@ -141,23 +196,26 @@ func (c *Client) Perform(req *http.Request) (*http.Response, error) {
...
-			nodeURL     *url.URL
+			conn        *Connection
 			shouldRetry bool
 		)
 
-		// Get URL from the Selector
+		// Get connection from the pool
 		//
-		nodeURL, err = c.getURL()
+		conn, err = c.getConnection()
```
urso: This makes me wonder if this is really a connection pool, or a clever URL selector. Using the stdlib Transport, we have the actual connections and connection pool handled in another layer. Plus: there is no limit of active connections, neither here, nor in the default stdlib pool.

estransport/connection.go
karmi: That's effectively hapenning with the "scheduled resurrect" — as soon as the connection is removed, it is not being returned from `Next()`, and the scheduler will try to "resurrect" it with an exponential delay.

estransport/connection.go
karmi: Yes, it's entirely possible that the user creates a goroutine per API call, and uses the client concurrently.

The `container/heap` package would be useful here, but I'm not sure if it would simplify things or confuse them... Will investigate at some later point.

estransport/estransport.go
karmi: Correct, this is basically a "smart URL selector". I went with the naming to keep it in line with the other official clients, since it's "on the fence" between a selector and connection pool.

estransport/connection.go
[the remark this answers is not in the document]
karmi: Correct, this interface could be used unexported just as well. As I've explained above, the plan is to make it possible for the user to provide their own implementation, and that's why I went with the exported types in certain parts, although they're not really useful at this point.

estransport/connection.go
urso: Yes, the round robin one has resurrection and retry support. But if only one URL is configured we will use singleConnectionPool, which doesn't seem to have some kind of backoff behavior.

karmi:
Thanks again for all the feedback — merged into `master`, `7.x` and `6.x` branches.

PR #97, opened 2019-10-25
Improve fasthttp example

erikdubbelboer:


karmi:
Thanks for the optimizations, @erikdubbelboer! I've tweaked the commit message and code annotations a little bit. I've also run the testing script and benchmarks locally, so the numbers are more representative.

PR #98, opened 2019-10-31
estransport: refactor Next/Resurrect

axw:
Refactor Next and Resurrect to share most of the code.
This fixes a bug whereby metrics would not have been updated when calling `Next`.


karmi:
Thanks, @axw!

PR #99, opened 2019-10-31
estransport: remove enableMetrics

axw:
These fields are redundant, as you can check that metrics is non-nil.

### DIFF
--- a/estransport/connection.go
+++ b/estransport/connection.go
@@ -40,8 +40,7 @@ type Connection struct {
 type singleConnectionPool struct {
 	connection *Connection
 
-	metrics       *metrics
-	enableMetrics bool
+	metrics *metrics
 
 	debugLogger DebuggingLogger
 }
@@ -54,8 +53,7 @@ type roundRobinConnectionPool struct {
 	dead []*Connection // List of dead connections
 	orig []*url.URL    // List of original URLs, passed in during initialization
 
-	metrics       *metrics
-	enableMetrics bool
+	metrics *metrics
 
 	debugLogger DebuggingLogger
 }
@@ -74,12 +72,7 @@ func newRoundRobinConnectionPool(u ...*url.URL) *roundRobinConnectionPool {
 
 	cp := roundRobinConnectionPool{live: conns, orig: u, curr: -1}
 
-	if cp.enableMetrics {
-		cp.metrics.Lock()
-		cp.metrics.live = cp.live
-		cp.metrics.dead = cp.dead
-		cp.metrics.Unlock()
-	}
+	// BUG the transport's metrics should be initialised with the live list
 
 	return &cp
 }
@@ -182,7 +175,7 @@ func (cp *roundRobinConnectionPool) Remove(c *Connection) error {
 	copy(cp.live[index:], cp.live[index+1:])
 	cp.live = cp.live[:len(cp.live)-1]
 
-	if cp.enableMetrics {
+	if cp.metrics != nil {
 		cp.metrics.Lock()
 		cp.metrics.dead = cp.dead
 		cp.metrics.live = cp.live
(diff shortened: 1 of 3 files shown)


estransport/connection.go
```diff
@@ -74,12 +72,7 @@ func newRoundRobinConnectionPool(u ...*url.URL) *roundRobinConnectionPool {
 
 	cp := roundRobinConnectionPool{live: conns, orig: u, curr: -1}
 
-	if cp.enableMetrics {
```
axw: As @simitt previously mentioned, `cp.enableMetrics` would always be false here; the code I've deleted is dead code.

If you want people to be able to implement `ConnectionPool` from outside this package, then I would suggest braking the dependency from the implementations back to the `metrics` type. Instead, have the pools implement some interface that returns connection pool metrics (which must be an exported type, as otherwise nobody outside can implement it...)

Also, I'd suggest making `Connection` and interface too, otherwise `ConnectionPool` implementations cannot specialise the connection according to the pool. For example, the `roundRobinConnectionPool` could create a connection object that tracks its position in the "live" and "dead" lists, so you don't have to scan through the list each time you move them. Alternatively, it could maintain pointers for use in a linked list.

So something like this:

```go

type ConnectionPool interface {
    Next() (Connection, error)
    Remove(Connection) error
    Metrics() ConnectionPoolMetrics
}

type ConnectionPoolMetrics struct {
    // (maybe this should also be an interface?)
}

type Connection interface {
    URL() *url.URL
    Info() ConnectionInfo
}

type ConnectionInfo struct {
    Dead bool
    DeadSince time.Time
    Failures int
}
```

Another option, which would lead to fewer interfaces, would be to have the pool only accept and return URLs. Everything else would be internal to the pool. e.g.

```go
type ConnectionPool interface {
    Next() (*url.URL, error)
    Metrics() ConnectionPoolMetrics

    // OnError is an alternative name for "Remove".
    OnError(*url.URL)

    // OnSuccess is an alternative name for "markAsHealthy", but
    // operating through the pool rather than directly on the connection object.
    OnSuccess(*url.URL)
}

type ConnectionPoolMetrics interface {
    String() string
}
```

Finally, you would update `Metrics`, removing its `Live` and `Dead` lists with a `Connections` field of type ConnectionPoolMetrics:

```go
type Metrics struct {
        Requests  int         `json:"requests"`
        Failures  int         `json:"failures"`
        Responses map[int]int `json:"responses"`
        Connections ConnectionPoolMetrics `json:"connections,omitempty"`
}
```

karmi: Thanks for the suggestions, @axw. Having the pool work only with `*url.URL` is something I was considering, but then I got worried about having to do lookups (in some sort of map) and other complications. However, I like how it looks. It's definitely something I would like to explore.

Side note, `Metrics()` is separate, defined by the `Measurable` interface, I'd like to keep it that way, if possible at all.

axw: > Side note, Metrics() is separate, defined by the Measurable interface, I'd like to keep it that way, if possible at all.

You could leave the Metrics method out of the ConnectionPool interface. My point was more that it needs to return metrics through an exported API, rather than the implementations updating the internal metrics type.

karmi: I've refactored the metrics support significantly in https://github.com/elastic/go-elasticsearch/commit/2642e1cf51b277e3f9a599fee2ab38ef7f70e0aa.

I've kept `Connections` as `[]*connectionMetric`, since I don't understand how working simply with a `fmt.Stringer` would work for the exported metrics, ie. converting them to JSON.

karmi: I guess I see your point now: allowing people to record/report custom information in the connection metric. I've simply made it a slice of `fmt.Stringer`:

https://github.com/elastic/go-elasticsearch/blob/980181bb7eefbd5fd9d26a97c70caaf1cac0afdc/estransport/metrics.go#L42

PR #126, opened 2020-01-29
Add the CACert configuration option and security example

karmi:
This patch adds a `CACert` configuration option to facilitate setting up the client for an environment with custom SSL certificates, and also adds executable examples to demonstrate the feature.


estransport/estransport.go
```diff
@@ -100,11 +104,31 @@ type Client struct {
...
 		cfg.Transport = http.DefaultTransport
 	}
 
+	if cfg.CACert != nil {
+		httpTransport, ok := cfg.Transport.(*http.Transport)
+		if !ok {
+			return nil, fmt.Errorf("unable to set CA certificate for transport of type %T", cfg.Transport)
+		}
+
+		sysCertPool, err := x509.SystemCertPool()
```
axw: Perhaps check with other clients, but I would strongly suggest using `x509.NewCertPool()` here, so the specified CA certs are the _only_ ones that are used for verification.

elasticsearch.go
```diff
@@ -37,6 +37,8 @@ type Config struct {
 	CloudID string // Endpoint for the Elastic Service (https://elastic.co/cloud).
 	APIKey  string // Base64-encoded token for authorization; if set, overrides username and password.
 
+	CACert []byte // PEM-encoded certificate authorities.
```
axw: I think you should also document behaviour with respect to system defaults: does it override or extend? (See further comments.)

Perhaps also mention that this only works if Transport is unspecified, or is an `*http.Transport`.

karmi: Addressed in dcdb7f2.

estransport/estransport.go
karmi: Addressed in dcdb7f2.

PR #127, opened 2020-01-31
[README] Add missing res.Body.Close()

denisvmedia:
This is a minor change to the readme file that adds a missing res.Body.Close().

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,6 +187,7 @@ func main() {
   if err != nil {
     log.Fatalf("Error getting response: %s", err)
   }
+  defer res.Body.Close()
   // Check response status
   if res.IsError() {
     log.Fatalf("Error: %s", res.String())


elasticcla:
Hi @denisvmedia, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in your Git [commit](https://github.com/elastic/go-elasticsearch/pull/127.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

karmi:
Hi, thanks for the catch! It looks like the commit has `d....v@home` as the email address, and that's why the check fails. Can you fix the mail in your Git configuration, and force push?

denisvmedia:
@karmi I fixed it, thanks :)

karmi:
Thanks for the patch! Merged into `master` and backported into `7.x` and `6.x`.

PR #131, opened 2020-02-10
Add index parameter to Indices.PutSettings

devbv:
Hello, 
I found missing parameter in Indices.PutSettings.
This patch makes api work as purpose.


karmi:
Hello, the API is generated from a specification, and the index name for "Put Settings" API is actually optional — you should use [`client.Indices.PutSettings.WithIndex()`](https://pkg.go.dev/github.com/elastic/go-elasticsearch/v7@v7.5.0/esapi?tab=doc#IndicesPutSettings.WithIndex) to pass the index name(s).

devbv:
@karmi Oh got it, thanks.

PR #137, opened 2020-02-29
Add the esutil.BulkIndexer component

karmi:
This pull request adds a parallel, asynchronous and efficient bulk helper.

The API is defined by the [**`esutil.BulkIndexer`**](https://github.com/elastic/go-elasticsearch/blob/bulk_helper/esutil/bulk_indexer.go#L26) interface:
https://github.com/elastic/go-elasticsearch/blob/d065e908b3c79d6fecb8387865c9b9a7c4f30e0b/esutil/bulk_indexer.go#L26-L41

The caller uses the `Add()` method to add a `esutil.BulkIndexerItem`, which is passed on to the workers, which encode the operation into the Bulk API format, and send it to Elasticsearch, when the size of the buffer passes the configured threshold. The indexer also performs a periodic flush, based on the configured interval.

The indexer accepts a configuration in the [**`esutil.BulkIndexerConfig`**](https://github.com/elastic/go-elasticsearch/blob/bulk_helper/esutil/bulk_indexer.go#L45) struct.

A simple, but realistic, example of using the indexer:

```golang
es, _ := elasticsearch.NewClient(elasticsearch.Config{
	RetryOnStatus: []int{502, 503, 504, 429},
	RetryBackoff: func(i int) time.Duration { return time.Duration(i) * 100 * time.Millisecond },
	MaxRetries: 5,
})

indexer, _ := esutil.NewBulkIndexer(esutil.BulkIndexerConfig{
	Client:        es,              // The Elasticsearch client
	Index:         "test",          // The default index name
	NumWorkers:    4,               // The number of worker goroutines (default: number of CPUs)
	FlushBytes:    2e+6,            // The flush threshold in bytes (default: 5MB)
	FlushInterval: 30 * time.Second // The flush interval (default: 10sec)
})

indexer.Add(
	context.Background(),
	esutil.BulkIndexerItem{
		Action: "index",
		Body: strings.NewReader(`{"title":"Test"}`),
	})

indexer.Close(context.Background())
```

The indexer provides **`OnSuccess()`** and **`OnFailure()`** callbacks for each item, to propagate the individual operation results to the caller. It also provides the **`Stats()`** method, returning the [**`esutil.BulkIndexerStats`**](https://github.com/elastic/go-elasticsearch/blob/bulk_helper/esutil/bulk_indexer.go#L73) struct for debugging and monitoring.

A commented, executable example is available in [**`_examples/bulk/indexer.go`**](https://github.com/elastic/go-elasticsearch/blob/bulk_helper/_examples/bulk/indexer.go):

```
$ go run indexer.go -count=100000 -flush=1000000

# BulkIndexer: documents [100,000] workers [8] flush [1.0 MB]
# ▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁
# → Generated 100,000 articles
# ▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔
# Sucessfuly indexed [100,000] documents in 1.909s (52,383 docs/sec)
```

A great deal of care has been taken to make the indexer as efficient as possible. A benchmark test is available in [**`esutil/bulk_indexer_benchmark_test.go`**](https://github.com/elastic/go-elasticsearch/blob/bulk_helper/esutil/bulk_indexer_benchmark_test.go):

```
$ go test -v -run=none -bench=BenchmarkBulkIndexer -race -benchmem -benchtime=10s .
BenchmarkBulkIndexer/Basic-4  1376808  8735 ns/op  178 B/op  3 allocs/op
```

The performance has been verified by end-to-end benchmarks against a real cluster, with both client and server running on separate Google Cloud Platform instances.

For a small document, the throughput is around 250,000 documents per second:

```
ELASTICSEARCH_URL=http://server:9200 go run benchmarks.go --dataset=small --count=1_000_000 --flush=2MB --shards=5 --replicas=0 --fasthttp=true --easyjson=true
small: run [10x] warmup [3x] shards [5] replicas [0] workers [8] flush [2.0 MB] wait [1s] fasthttp easyjson
▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔
docs/sec: min [279,173] max [289,351] mean [286,987]
```

For a document representing the Nginx log event captured by Filebeat, the throughput is around 50,000 documents per second:

```
ELASTICSEARCH_URL=http://server:9200 go run benchmarks.go --dataset=httplog --count=1_000_000 --flush=3MB --shards=5 --replicas=0 --fasthttp=true --easyjson=true
httplog: run [10x] warmup [3x] shards [5] replicas [0] workers [8] flush [3.0 MB] wait [1s] fasthttp easyjson
docs/sec: min [50,165] max [53,072] mean [52,011]
```

The "theoretical" performance of the indexer has been verified by running against an Nginx server, with a throughput surpassing 1,000,000 documents per second:

```
ELASTICSEARCH_URL=http://server:8000 go run benchmarks.go --dataset=small --count=1_000_000 --flush=2MB --warmup=0 --mockserver
small: run [10x] warmup [0x] shards [3] replicas [0] workers [8] flush [2.0 MB] wait [1s]
docs/sec: min [1,187,648] max [1,253,132] mean [1,233,818]
```

Please see [**`_examples/bulk/benchmarks`**](https://github.com/elastic/go-elasticsearch/tree/bulk_helper/_examples/bulk/benchmarks) for more detailed information about the performance, the testing environment used, and run the script to see percentile statistics, omitted from the output here.

Note that the benchmarks use the `valyala/fasthttp` package for improving the performance of HTTP transport, and a custom JSON decoder provided by the `mailru/easyjson` package.

-----

/cc @elastic/es-clients @axw @urso @simitt @tonymeehan


esutil/bulk_indexer.go
```diff
@@ -0,0 +1,509 @@
...
+			if err := w.writeBody(item); err != nil {
+				if item.OnFailure != nil {
+					item.OnFailure(item, BulkIndexerResponseItem{}, err)
+				}
+				atomic.AddUint64(&w.bi.stats.numFailed, 1)
+				w.mu.Unlock()
+				continue
+			}
+
+			w.items = append(w.items, item)
```
axw: It would to avoid appending to `w.items` if there's no callback(s) defined. That would require you to record the index on the items, for matching up with the results.

esutil/bulk_indexer.go
```diff
@@ -0,0 +1,509 @@
...
+
+	Metadata interface{} // To use eg. in OnSuccess/OnFailure callbacks
+
+	OnSuccess func(item BulkIndexerItem, res BulkIndexerResponseItem)            // Per item
+	OnFailure func(item BulkIndexerItem, res BulkIndexerResponseItem, err error) // Per item
+}
+
+// BulkIndexerResponse represents the Elasticsearch response.
+//
+type BulkIndexerResponse struct {
```
axw: Another pro for removing configurable decoder: this type could be unexported/removed, making it possible to decode "items" in a streaming fashion (not necessarily immediately, but at some point).

esutil/bulk_indexer.go
```diff
@@ -0,0 +1,509 @@
...
+
+// Close calls flush on writers and closes the indexer queue channel.
+//
+func (bi *bulkIndexer) Close(ctx context.Context) error {
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	default:
+		close(bi.queue)
+		bi.wg.Wait()
```
axw: This waits for an arbitrary amount of time, so the context timeout/cancellation may not be effective. Perhaps use a channel to signal when all workers are closed, and select on it along with ctx.Done.

esutil/bulk_indexer.go
```diff
@@ -0,0 +1,509 @@
...
+	SourceExcludes      []string
+	SourceIncludes      []string
+	Timeout             time.Duration
+	WaitForActiveShards string
+}
+
+// BulkIndexerStats represents the indexer statistics.
+//
+type BulkIndexerStats struct {
+	NumAdded    uint
```
axw: What's the reason for using `uint`, when internally it's `uint64`? May as well be consistent I think, and avoid possible truncation.

esutil/bulk_indexer.go
```diff
@@ -0,0 +1,509 @@
...
+//
+type BulkIndexer interface {
+	// Add adds an item to the indexer. It returns an error when the item cannot be added.
+	// Use the OnSuccess and OnFailure callbacks to get the operation result for the item.
+	//
+	// You must call the Close() method after you're done adding items.
+	//
+	// It is safe for concurrent use. When it's called from goroutines,
+	// they must finish before the call to Close, eg. using sync.WaitGroup.
+	Add(context.Context, BulkIndexerItem) error
```
axw: Did you consider exposing the channel, so that users could select on it? e.g.

```go
type BulkIndexer struct {
    C() chan<- BulkIndexerItem
    Close(context.Context) error
    Stats() BulkIndexerStats
}
```

esutil/bulk_indexer.go
```diff
@@ -0,0 +1,509 @@
...
+				if bi.config.DebugLogger != nil {
+					bi.config.DebugLogger.Printf("[indexer] Auto-flushing workers after %s\n", bi.config.FlushInterval)
+				}
+				for _, w := range bi.workers {
+					w.mu.Lock()
+					if w.buf.Len() > 0 {
+						if err := w.flush(); err != nil {
+							// TODO(karmi): Wrap error
+							w.mu.Unlock()
+							return fmt.Errorf("%s", err)
```
axw: This is going to cause the periodic flusher to exit on the first error - is that desirable? What about transient errors, like a network outage? Probably not great behaviour for a long-lived bulk indexer.

esutil/bulk_indexer.go
```diff
@@ -0,0 +1,509 @@
...
+	case <-ctx.Done():
+		return ctx.Err()
+	default:
+		close(bi.queue)
+		bi.wg.Wait()
+	}
+
+	bi.ticker.Stop()
+
+	for _, w := range bi.workers {
```
axw: Rather than flushing here, how about having the workers flush upon seeing their channel is closed, but before signalling that they're done?

esutil/bulk_indexer.go
```diff
@@ -0,0 +1,509 @@
...
+	if res.Body != nil {
+		defer res.Body.Close()
+	}
+	if res.IsError() {
+		atomic.AddUint64(&w.bi.stats.numFailed, uint64(len(w.items)))
+		// TODO(karmi): Wrap error (include response struct)
+		return fmt.Errorf("flush: %s", res.String())
+	}
+
+	if err := w.bi.config.Decoder.UnmarshalFromReader(res.Body, &blk); err != nil {
```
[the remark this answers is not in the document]
karmi: I see a lot of value in exposing a configurable JSON decoder; let me explain the various points:

* As of now, the package is [completely dependency-free](https://github.com/elastic/go-elasticsearch/blob/master/go.mod), with only the stdlib packages being imported. I would try my best to keep it that way. (Notice the demonstrations in `_examples` use a lot of third-party packages, but they are not part of the package itself.) Of course, this is not a dogma, but I haven't yet found a good-enough reason to break this commitment. In other words, it doesn't feel right to me to add a third-party dependency to the entire package for something such as parsing an HTTP response in the `esutil` sub-package.

* Furthermore, I see a configurable decoder as a fundamental feature of the package as a whole. For context, it did serve us very good for the original client architectures in Ruby and Python — a configurable `Serializer` component is something a lot of people find use for. (See eg. [this example](https://gist.github.com/lribeiro/7d9fbedf830a54685811f63dbbce9464) for implementing a custom serializer for Python's `numpy` library).

* I expect to use a strategy like this for any other utility component which has to deal with JSON encoding/decoding, like a possible `Search()` or `Scan()`  helper.

esutil/bulk_indexer.go
axw: Fair enough, and I agree - just wondering if you considered this option, as I brought it up elsewhere. I think selecting on add would _probably_ be unusual anyway, so it isn't so relevant here.

esutil/bulk_indexer.go
```diff
@@ -0,0 +1,509 @@
...
+		}
+		w.buf.WriteRune('\n')
+	}
+	return nil
+}
+
+// flush writes out the worker buffer; it must be called under a lock.
+//
+func (w *worker) flush() error {
+	if w.isFlushing {
```
simitt: I might be overlooking something, but when the callers need to safeguard that `flush` is called under a lock, then how could it happen that it is called while flushing? And if it can happen, is it then safe that there is no explicit lock around reading and writing `w.isFlushing`? 
As far as I understand, the `isFlusing` can be completely dismissed.

esutil/bulk_indexer.go
axw: No, it wouldn't stop the entire indexer - but the periodic flusher would no longer operate. Seems wrong to me to leave the indexer in a half-working state.

esutil/bulk_indexer.go
```diff
@@ -0,0 +1,509 @@
...
+			info BulkIndexerResponseItem
+			op   string
+		)
+
+		item = w.items[i]
+		for k, v := range blkItem {
+			op = k
+			info = v
+		}
+		if info.Error.Type != "" || info.Status > 201 {
```
simitt: Why are you using `201` here? The `BulkeRequest.Do` returns `true` for `response.isError()` when `statusCode >= 299` which I think is more correct.

esutil/bulk_indexer.go
axw: Sure, this doesn't need to be done now. I'll stick to reviewing things that affect the external API then.

esutil/bulk_indexer.go
```diff
@@ -0,0 +1,509 @@
...
+			item BulkIndexerItem
+			info BulkIndexerResponseItem
+			op   string
+		)
+
+		item = w.items[i]
+		for k, v := range blkItem {
+			op = k
+			info = v
+		}
```
simitt: I am confused by this range loop, shouldn't that contain the code below? The way it is now it loops over the `blkItem`which is a `map[string]BulkIndexerResponseItem`, and sets `op` and `info`; maps don't have a defined order, so it is undefined what `op` and `info` is set to after the loop, and only the last entry will be processed further.
It implicitly works as there is always only one entry in the map per `blkItem`, but this is hard to read and could easily lead to bugs if the response object was created in a different way. If it is intended to always only contain one item and you want to add a label to the `BulkIndexerResponseItem` I suggest to create a dedicated struct holding those two pieces of information rather than creating a map.

esutil/bulk_indexer.go
karmi: Please keep raising the questions, though! Even though all the suggestions might not be implemented right away, it's good to keep track of them.

esutil/bulk_indexer.go
```diff
@@ -0,0 +1,509 @@
...
+	Index           string
+	Action          string
+	DocumentID      string
+	Body            io.Reader
+	RetryOnConflict *int
+
+	Metadata interface{} // To use eg. in OnSuccess/OnFailure callbacks
+
+	OnSuccess func(item BulkIndexerItem, res BulkIndexerResponseItem)            // Per item
+	OnFailure func(item BulkIndexerItem, res BulkIndexerResponseItem, err error) // Per item
```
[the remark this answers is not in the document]
axw: Isn't that what I suggested? i.e. the BulkIndexer accepts a channel through the BulkIndexerItem. However, it's not unbounded: it's 0 or 1.

esutil/bulk_indexer.go
axw: I would make it uint64, as otherwise on 32-bit systems you'll risk truncating the metrics and getting weird inconsistencies.

esutil/bulk_indexer.go
axw: This is due to the [_bulk](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html) result format. For each item, you get back a map with a single key/value pair, where the key is the operation name.

esutil/bulk_indexer.go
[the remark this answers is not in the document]
karmi: Ah right, sorry, I misread the code.

I was thinking about it a little bit more, and I'm leaning towards keeping the notifications as callbacks. In a talk linked from the article cited above, Alan Schreve makes a good point — at around [16:50](https://youtu.be/hFqXgmor74k?t=1010) —, that a callback-type of notification can be easily turned into a channel-based notification, but not vice versa. I'm thinking about something like:

```golang
func(item BulkIndexerItem, res BulkIndexerResponseItem) {
	updatedIDs <- item.DocumentID
}
```

I've played with it a little bit in the unit tests, pushed to a separate branch here: https://github.com/elastic/go-elasticsearch/commit/706ad08e730101d898fdc39beb5f7aaf4abae352.

esutil/bulk_indexer.go
karmi: Yes, correct — thanks for the explanation.

esutil/bulk_indexer.go
karmi: That is correct — thanks for raising it, it's a buggy behaviour.

esutil/bulk_indexer.go
simitt: Ah, I overlooked that this is due to the ES bulk response, that makes sense then. Resolving this conversation.

esutil/bulk_indexer.go
karmi: Yeah, I'll need to look into it again — previously, there was no mutex synchronization, before I added the periodic flush interval, so this might only a residue of the previous implementation. (Note the flush is called from `Close()`.)

esutil/bulk_indexer.go
karmi: > It implicitly works as there is always only one entry in the map per blkItem (..)

The Bulk API response is a map with only a single key, and I don't think that can change, since it corresponds to the operation, ie. for a "create" operation, you'll get back a "create" key. So it is kind of "contractually correct" :) I agree the code is a bit too hard-to-read here. I'll think about this some more.

esutil/bulk_indexer.go
karmi: Note that this is a status _for item_, part of the JSON payload in the response body, not the overall HTTP response — for a created document, the status code will be `201`.

esutil/bulk_indexer.go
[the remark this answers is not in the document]
karmi: Right, I see. I'm not against what you're suggesting here, but I guess I'm just not sure how to implement the JSON parsing for this well-defined structure without explicitely — or implicitely — relying on a third-party package. Do you have a suggestion for that?

In general, JSON parsing by the standard library _is_ quite a severe bottleneck — I would have to run the benchmarks again to be able to provide some numbers, but I did see a drop in performance there.

Is your concern mainly about exposing a part of the public API — the `Decoder` component —, unneccessarily?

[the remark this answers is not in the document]
axw: > Right, I see. I'm not against what you're suggesting here, but I guess I'm just not sure how to implement the JSON parsing for this well-defined structure without explicitely — or implicitely — relying on a third-party package. Do you have a suggestion for that?

Not yet. It'll probably involve reinventing the wheel - i.e. yet another decoder generator, but one that generates all the required code. How feasible that is I'm not yet sure.

> Is your concern mainly about exposing a part of the public API — the Decoder component —, unneccessarily?

Yes.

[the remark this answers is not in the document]
karmi: I see, thanks for the clarification.

I don't think that we can reinvent the wheel in a matter of days :) Because I wouldn't like to ship this feature with a known performance degradation, coming from `encoding/json`, I would rather the `Decoder` extension point as is, and deprecate, or even straight out remove it later.

esutil/bulk_indexer.go
urso: 💯 for not using go-channels. I can always wrap the Indexer in a go-routine and use channels for the callback, but working around APIs that enforce go-channels can be a pain.

esutil/bulk_indexer.go
```diff
@@ -0,0 +1,509 @@
...
+	// Use the OnSuccess and OnFailure callbacks to get the operation result for the item.
+	//
+	// You must call the Close() method after you're done adding items.
+	//
+	// It is safe for concurrent use. When it's called from goroutines,
+	// they must finish before the call to Close, eg. using sync.WaitGroup.
+	Add(context.Context, BulkIndexerItem) error
+
+	// Close waits until all added items are flushed and closes the indexer.
+	Close(context.Context) error
```
urso: I wonder if the behavior of close overlaps with context cancellation and wait groups. Do I need to cancel added items if I want close to return early? If cancelled, will my callback be run?

Maybe it would make sense to have 'Close' on the concrete type only, but remove it from the interface?

esutil/bulk_indexer.go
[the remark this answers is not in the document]
urso: I'd prefer to have a callback. As @karmi said, I can easily add channels with semantics I need. E.g. 
- a channel could be global on the Indexer
- two per global Indexer (success, failure) reporting the IDs that have been indexed or failed
- one per Item. If per item do I have to create it and the API will close it for me? If so I can easily do this via callbacks without enforcing a 'signaling' protocol between user and API.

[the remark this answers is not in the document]
urso: Have you considered one single callback like `OnProcessed(BulkIndexerItem, BulkIndexerResponseItem)` (with the `BulkIndexerResponseItem` also including an error value)? This reduces the number of callbacks one has to implement (at the potential cost of an if-else).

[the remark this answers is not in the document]
karmi: Thanks for the comment, @urso! The more I think about it, the more I think the callback-based approach is both safer for the calling code, as well as more flexible (as outlined above). But I'm open to further discussion, of course.

esutil/bulk_indexer.go
```diff
@@ -0,0 +1,509 @@
...
+//
+type BulkResponseJSONDecoder interface {
+	UnmarshalFromReader(io.Reader, *BulkIndexerResponse) error
+}
+
+// BulkIndexerDebugLogger defines the interface for a debugging logger.
+//
+type BulkIndexerDebugLogger interface {
+	Printf(string, ...interface{})
+}
```
urso: We have this problem with our Kafka library. What is debug and what is info we want to show (some state change due to sniffing for example). Just seeing Printf makes me wonder if every log is really 'debug'.

As we have our own interface, would it make sense to have an interface (this one or an additional one) that gives away a little more information for the purpose of structured logging? E.g. if there was an IO error, which host?

esutil/bulk_indexer.go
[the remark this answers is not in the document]
karmi: > Have you considered one single callback like `OnProcessed` (...)

That's an interesting idea, to combine everything into a single "handler". I can see how this would simplify the calling code, but, as you note, it would have to need to differentiate between success and failure inside the callback.

This could get rather complex, like checking for `err != nil`, then checking if `res.IsError()`, and so on... But there's something intriguing about an API like that — I'll think about it some more.

esutil/bulk_indexer.go
```diff
@@ -0,0 +1,509 @@
...
+			Reason string `json:"reason"`
+		} `json:"caused_by"`
+	} `json:"error,omitempty"`
+}
+
+// BulkResponseJSONDecoder defines the interface for custom JSON decoders.
+//
+type BulkResponseJSONDecoder interface {
+	UnmarshalFromReader(io.Reader, *BulkIndexerResponse) error
+}
```
urso: I'm not sure I would expose the decoder. The structure and it's struct tags are cleary 'biased' towards the stdlib json decoder. Not all JSON libraries do support all struct tags. Instead we might consider to choose an efficient decoder only and return a struct without json tags.

Appropos fast json parser, have you seen: https://github.com/minio/simdjson-go

esutil/bulk_indexer.go
urso: Yeah, the `ctx.Done()` is somewhat ineffective. Unfortunately Wait does not mix well with channels. We also might want to protect the indexer from being closed mutliple times. Alternative pattern:

```
bi.closeOnce.Do(func() { close(bi.queue) })
chWait := make(chan struct {})
go func() { bi.wg.Wait(); close(chWait) }
select {
case <-ctx.Done():
  return ctx>Err()
case <- chWait:
  return nil
}
```

If the bulkIndexer is still busy shutting down, then we leak a go-routine.

Is there also an option to discard all events before closing? In case we don't want to wait?

esutil/bulk_indexer.go
karmi: That's a good point. This "logger", however, is really something only relevant for debugging — effectively, it just wrapped the `fmt.Printf` statements I have sprinkled over the codebase during development.

As there's no common logging interface in Go, it would be rather hard to support something like a structured logging — I've struggled with that couple of times already.

I'm thinking that instead of having the component doing the logging, the _calling code_ should do the logging itself, by using some kind of hooks. This is easy for events like "error indexing item", by using the `OnFailure()` callback. But for more low-level events, like "worker 001 started", we would need to provide specific hooks for them. I'm not sure if low-level events like these are terribly useful for a production system, though?

Note that the component exposes the `BulkIndexerStats`, which is something I imagine a logging/monitoring system would consume...

esutil/bulk_indexer.go
```diff
@@ -0,0 +1,509 @@
...
+			ch:  bi.queue,
+			bi:  bi,
+			buf: bytes.NewBuffer(make([]byte, 0, bi.config.FlushBytes)),
+			aux: make([]byte, 0, 512)}
+		w.run()
+		bi.workers = append(bi.workers, &w)
+	}
+	bi.wg.Add(bi.config.NumWorkers)
+
+	bi.ticker = time.NewTicker(bi.config.FlushInterval)
```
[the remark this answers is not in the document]
urso: +1 for adding the flush timer to the workers. The workers do the buffering already, which means that they are somewhat 'independent' already.

esutil/bulk_indexer.go
urso: moving flush to the worker would help/solve this.

esutil/bulk_indexer.go
urso: Nevermind. Just seeing the discussion about the decoder, not relying on third party libs, and possible performance bottleneck with stdlib json.

esutil/bulk_indexer.go
urso: The items will be stored on heap. Due to the flush you reuse the items array, making it a rather cheap operation (no extra allocation). Drawback is that you have no upper bound on the size of the array. The size is implicitely define to be the maximum number of Events you have seen in the past before you to reach the flush limit.

I presume having an items array makes it easier to handle the response, as the bulk api response contains the per item information in the same order (and same number of entries) as this array has. By only storing items that have a callback one might be forced to correlate item in the array with items in the response.

esutil/bulk_indexer.go
urso: Yeah, I'm hinting at introducing hooks. At least for very important events. But I guess the more important events (e.g. sniffing failed) will be comming from the esclient lib itself.

esutil/bulk_indexer.go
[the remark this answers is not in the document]
axw: I know we can wrap a channel with functions. I'm asking the question to get a feel for what we're optimising for, e.g. do the callers of "Add" typically want to wait for the bulk response? If so, a channel will be most useful. For other cases where you want an observer style interface where you're notified of all responses, then I think a callback would be fine.

That is to say, I think there's room for two mechanisms:
 - a callback/interface configured on the BulkIndexer itself, which is invoked for each request/response
 - a callback/interface/channel on the item, which is invoked on that item's response

The first one would be useful for things like metrics, which won't significantly vary the callback behaviour based on the item.

It might also be worth considering an interface, rather than choosing either a function or channel. Implementations of both could be provided. e.g.

```go
type BulkResponse struct {
    Item BulkIndexerItem
    Err error // If Err is non-nil, Response is undefined
    Response BulkIndexerResponseItem
}

type BulkResponseHandler interface {
    HandleBulkResponse(BulkResponse)
}

type BulkResponseHandlerFunc func(BulkResponse)
func (f BulkResponseHandlerFunc) HandleBulkResponse(...) {...}

type BulkResponseHandlerChan chan<- BulkResponse
func (ch BulkResponseHandlerChan) HandleBulkResponse(...) {...}
```

Just a thought.

+1 to combining OnSuccess/OnFailure anyway.

esutil/bulk_indexer.go
karmi: Yes, the more important events will happen on the client level. I'm thinking about removing the `DebugLogger` altogether, to prevent the public API pollution. 

It would be indeed a nice feature to add hooks, which the user can implement and pass to configuration — I'm just not sure I want to cram all of that into a first pull request...

esutil/bulk_indexer.go
karmi: I went through several different names, including `Wait()`, but finally decided on `Close()`. It does not "close" the indexer in a sense of "drop what you're doing and shut down", but rather signals to the indexer that it should finish the in-flight operations and close itself for more.

> Do I need to cancel added items if I want close to return early?

Well, currently there's no semantics of returning early. As `Close()` takes a context, I imagine that  propagating eg. a timeout to the `worker.flush()` method would be possible — but I have focused only on making it possible, not implementing it.

> If cancelled, will my callback be run?

Currently, the callback should be called for every added item — also in case of cancel/timeout... But I would have to write some code to test it.

esutil/bulk_indexer.go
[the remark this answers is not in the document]
karmi: > I'm asking the question to get a feel for what we're optimising for, e.g. do the callers of "Add" typically want to wait for the bulk response?

I would expect most callers are interested in an answer to the question "what has happened to my item", most commonly creating a document. In case that operation wasn't successful, they would want to put that item into some sort of dead-letter queue, for inspection and re-processing.

The details of the HTTP response itself (ie. the result of a request to `_bulk`) is not something terribly interesting to the caller, I imagine. That's why I have designed the callbacks to be per-item since the start, differently from the Java `BulkProcessor` class or its Go port in `olivere/elastic`. (I had callbacks for these events as well, `OnBatchSuccess()` etc, but removed them during development.)

> It might also be worth considering an interface (...)

In general, I'm a big proponent of putting an implementation behind an interface. In this case, however, I feel like it introduces too much ceremony around a simple `func`-based callback. With the `func`-based callback, the functionality is clear from reading a godoc output or scanning the source code. Also, those additional types add more weight to an already big public API, and since this component lives in the rather generic `esutil` package, it might become quite big. (I've considered putting the component into a subpackage, something like `esbulk`, but wasn't sure if it wouldn't impact the discoverability of the component too much. The new [godoc website](https://pkg.go.dev/github.com/elastic/go-elasticsearch/v7?tab=doc) makes it arguably even harder to discover "Subdirectories"...)

> +1 to combining OnSuccess/OnFailure anyway.

Can you clarify why would you do this? As mentioned above, this shifts the responsibility of detecting the result of the operation to the caller. Sure, it's "just an `if/else`", mostly. But to me, it's more about _defining the semantics_ of what constitutes a success and error. Consider this code (which Silvia has asked about):

https://github.com/elastic/go-elasticsearch/blob/6265c4b76aee00190e6d7de45495e69cb39cafaa/esutil/bulk_indexer.go#L477

This is handling eg. this result in the Bulk API response, which indicates that a document to be deleted wasn't found:

https://github.com/elastic/go-elasticsearch/blob/6265c4b76aee00190e6d7de45495e69cb39cafaa/esutil/testdata/bulk_response_1b.json#L22-L34

Notice that the response JSON has `"errors": false` in the top-level field, and that there's no `error` for the individual item. However, deleting a document which wasn't found constitutes an error to me, and I would expect the `OnFailure()` callback to fire. This type of knowledge requires deep understanding how Elasticsearch behaves, and it took myself some time to understand what's going on, and a lot of confusion about why this operation isn't marked with an `error`. Therefore, I would like to depend on a high-level component such as this one to make these informed decisions.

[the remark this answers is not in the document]
urso: > Can you clarify why would you do this? As mentioned above, this shifts the responsibility of detecting the result of the operation to the caller. Sure, it's "just an if/else", mostly

If having to write callbacks, personally I'd prefer to write only one callback with if-else instead of two separate callbacks.

Using interfaces is a nice idea, in case we want to have multiple callbacks.

For both, the callback approach (having a single callback) and the interfaces approach we can introduce constructors wrapping different use-cases. E.g.

```
ch := make(chan struct{})
...

BulkIndexerResponseItem{
  ...
  OnProcessed: esutil.Notify(ch)
}
```

with:

```
func Notify(ch chan struct{}) BulkResponder {
  return func(...) {
    close(ch)
  }
}
```

I'd say having a function is not that much different from an interface, as in either case we might want to provide additional helpers.

esutil/bulk_indexer.go
urso: I see. I think some early cancel + shutdown is a must. If the library is used by a daemon, then daemon restart or shutdown should not be blocked by an unresponsive network or Elasticsearch instance.

Having a context.Context on Close for cancellation just feels a little weird.

esutil/bulk_indexer.go
urso: In case it's only debug, I'd still prefer to have a Printf style hook over no hook. At least it gives me some logs I can have a look at in case things go wrong.

esutil/bulk_indexer.go
karmi: Right right, thank you both. I'll have a proper look into the `Close()` and `Add()` behaviour, also because right now an `Add()` after `Close()` would panic, which is not nice.

esutil/bulk_indexer.go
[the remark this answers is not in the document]
karmi: Thanks for all the thoughts. At the moment, I'm more or less on the side of the fence where:

* `func`-style callbacks are used for notification,
* there are two separate callbacks, for success and failure.

Regarding the separate callbacks, I can't force myself to merge them into one... Among conceptual problems, I struggle with the naming. `OnProcessed()` seems wrong to me — if there's an error, the item hasn't really been "processed" at all.

Also, I've added a `BulkIndexerConfig.OnError()` callback in 9ec0d6b, to properly propagate errors from the depth of the indexer, and notify the calling code. Can you please look at that?

esutil/bulk_indexer.go
karmi: This had been removed in 9ec0d6b: 
https://github.com/elastic/go-elasticsearch/blob/9ec0d6bcffba0fe59174febe0e7c15ae8d58521f/esutil/bulk_indexer.go#L294-L300

esutil/bulk_indexer.go
karmi: Right! I'll have a look into propagating the errors, now that it's been made possible by 9ec0d6b. As @axw suggested above, `flush()` must take a context, and then it should be possible to pass down the context with timeout from `Close()` into it.

esutil/bulk_indexer.go
```diff
@@ -0,0 +1,535 @@
...
+	}
+
+	bi.init()
+
+	return &bi, nil
+}
+
+// Add adds an item to the indexer.
+//
+func (bi *bulkIndexer) Add(ctx context.Context, item BulkIndexerItem) error {
```
karmi: @axw, one thing I haven't yet figured out is how to properly instrument the underlying client with this interface. It's nice that `Add()` takes a context, but there's no way to propagate that to the `flush()` and `BulkRequest.Do()` down the line at the moment. One solution is to add an unexported field to `BulkIndexerItem` — the way the client does it —, and set it here, in this method. Do you have other ideas / suggestions?

axw: @karmi Elastic APM doesn't currently support assigning multiple parents to a transaction or span, so it wouldn't really work to propagate trace context through Add.

If you want to trace the bulk requests, I'd suggest instead creating transactions in the flusher. I'm beginning to think there should be a Flusher interface on the config, with a default implementation exported.

esutil/bulk_indexer.go
```diff
@@ -0,0 +1,535 @@
...
+type BulkIndexerConfig struct {
+	NumWorkers    int           // The number of workers. Defaults to runtime.NumCPU().
+	FlushBytes    int           // The flush threshold in bytes. Defaults to 5MB.
+	FlushInterval time.Duration // The flush threshold as duration. Defaults to 10sec.
+
+	Client      *elasticsearch.Client   // The Elasticsearch client.
+	Decoder     BulkResponseJSONDecoder // A custom JSON decoder.
+	DebugLogger BulkIndexerDebugLogger  // An optional logger for debugging.
+
+	OnError func(error) // Called for indexer errors.
```
axw: Seeing as you have no context around what part of the indexer is emitting an error, what can you do with the error other than log it? This doesn't seem particularly useful to me - I'd probably just extend the logger to also log errors, or perhaps just log them as debug messages.

On the other hand, it might be helpful to know when a _bulk request has failed overall, or how long bulk requests are taking, etc. I think it might be useful to add another interface like follows to BulkIndexerConfig:

```go
type BulkIndexerFlusher interface {
    // Writer is used for writing the bulk operations.
    io.Writer

    // Flush sends a bulk request with the written bulk operations,
    // and then clears the buffer.
    Flush(context.Context) (*esapi.Response, error)
}

type DefaultBulkIndexerFlusher struct {
    Client *elasticsearch.Client
  
    buf bytes.Buffer

    // Parameters of the Bulk API.
    Index               string
    ErrorTrace          bool
    FilterPath          []string
    Header              http.Header
    Human               bool
    Pipeline            string
    Pretty              bool
    Refresh             string
    Routing             string
    Source              []string
    SourceExcludes      []string
    SourceIncludes      []string
    Timeout             time.Duration
    WaitForActiveShards string
}

func (f *DefaultBulkIndexerFlusher) Write(data []byte) (int, error) {
    return f.buf.Write(data)
}

func (f *DefaultBulkIndexerFlusher) ReadFrom(r io.Reader) (int64, error) {
    // Implement io.ReaderFrom to more efficient io.Copy, when buffering item bodies.
    return f.buf.ReadFrom(r)
}

func (f *DefaultBulkIndexerFlusher) Flush(ctx context.Context) (*esapi.Bulk, error) {
    return f.Client.Bulk(f.buf, f.Client.Bulk.WithIndex(f.Index), f.Client.WithContext(ctx))
}
```

With this approach you could wrap the flusher for instrumentation purposes. The flusher would be invoked from workers. Workers would take on the responsibility of decoding the response, and dispatching the callbacks.

karmi: We have discussed this topic with @axw earlier today, to clarify some questions I had about the change. I've been thinking about it more during the day, and will summarize my thoughts.

One thing I don't like about this change is that it effectively moves the Bulk API parameters from the main configuration, `esutil.BulkIndexerConfig`, into the `Flusher`. For anybody who would like to set eg. an index name — which I assume is most of the users of the component —, or a pipeline — again, a frequent use case —, this would mean dealing with those in the custom component.

Another thing which bothers me is that I don't feel like restructuring the codebase significantly, ie. moving the buffer operations inside the `Flusher` (as I understand the proposal suggests). As I've said in earlier comments, I don't feel like delivering _everything and anything_ at once is a good development strategy. I would rather focus on the public API, so we don't have to break it in three weeks when we discover we need to eg. add a parameter to a method, and postponed _expanding_ the API later (which wouldn't break backwards compatibility).

I took a stab at extracting the basic operation needed for instrumentation, ie. the Bulk API call itself, into a separate component, and came up with the implementation in https://github.com/elastic/go-elasticsearch/commit/6a73580c2184f14386cf558517f2829116de3268. To me, this seems like solving the problem of wrapping this operation for instrumentation purposes. Can you please check that, @axw?

I've added an Elastic APM instrumentation into the Kafka demo in https://github.com/elastic/go-elasticsearch/commit/4ee74bbd1616faecc5e404c393a5bc1b011307f2, to verify how it behaves in reality. I've observed transactions being correctly recorded. I have not been successful with trying to capture errors: I've simulated many different types of errors — eg. sending an empty body in the bulk payload, which triggers and Elasticsearch error —, but the _Errors_ tab in the APM UI is empty. Maybe I'm doing something wrong – would you be able to check that as well, @axw?

axw: > One thing I don't like about this change is that it effectively moves the Bulk API parameters from the main configuration, esutil.BulkIndexerConfig, into the Flusher. For anybody who would like to set eg. an index name — which I assume is most of the users of the component —, or a pipeline — again, a frequent use case —, this would mean dealing with those in the custom component.

That change was based on the Flusher creating the request. With your approach of having the worker create the request and hand it off to the Flusher, it's fine to keep those params in the BulkIndexerConfig. With one exception: the client.

In your example: https://github.com/elastic/go-elasticsearch/commit/4ee74bbd1616faecc5e404c393a5bc1b011307f2#diff-7c873b1d59a93f95bb9ecbcc9a91c8baR103

You're passing in an elasticsearch.Client, but also passing it into the custom Flusher. The first one is not used at all in that case, which I find confusing. It would be simpler if you said: if BulkIndexerConfig.Flusher is nil, then a BulkIndexerDefaultFlusher will be constructed with a elasticsearch.NewDefaultClient. If you need to customise it, set Flusher. Then it becomes:

```go
esutil.NewBulkIndexer(esutil.BulkIndexerConfig{
    Flusher: &BulkIndexerDefaultFlusher{Client: client},
})
```

> Another thing which bothers me is that I don't feel like restructuring the codebase significantly, ie. moving the buffer operations inside the Flusher (as I understand the proposal suggests). As I've said in earlier comments, I don't feel like delivering everything and anything at once is a good development strategy. I would rather focus on the public API, so we don't have to break it in three weeks when we discover we need to eg. add a parameter to a method, and postponed expanding the API later (which wouldn't break backwards compatibility).

This directly affects the API: where the buffer sits changes the roles and responsibilities. If it's in the worker, then you don't need the io.Writer interface on the Flusher. If it's in the Flusher, you do. I think keeping it in the worker is fine though.

> I've added an Elastic APM instrumentation into the Kafka demo in 4ee74bb, to verify how it behaves in reality. I've observed transactions being correctly recorded. I have not been successful with trying to capture errors: I've simulated many different types of errors — eg. sending an empty body in the bulk payload, which triggers and Elasticsearch error —, but the Errors tab in the APM UI is empty. Maybe I'm doing something wrong – would you be able to check that as well, @axw?

Because you're using `apm.CaptureError(context, error)`, which only works if the supplied context contains a transaction or span. If you want to send errors without any transaction/span, then you'll need to use `apm.DefaultTracer.NewError(err).Send()` instead.

This is what I was getting at before (offline), about the OnError function being separated from the calling context. If you were capturing the error in the InstrumentedFlusher, then you would use the context returned by apm.ContextWithTransaction, and it would work nicely.

esutil/bulk_indexer.go
karmi: Thanks, this is indeed not needed, removed in a129cd4!

esutil/bulk_indexer.go
karmi: I've added a comment to make it more clear for future readers in 0212193.

esutil/bulk_indexer.go
karmi: We have collaborated on this with @axw earlier today, and ultimately solved the issue in a different way; the `Flusher` interface added yestarday has been removed from the codebase.

Two new callbacks have been added to the configuration in https://github.com/elastic/go-elasticsearch/commit/aa612f9daef5f04549ab8eca251cdf20eeb755f7:

```
OnFlushStart func(context.Context) context.Context
OnFlushEnd   func(context.Context)
```

This allows to manipulate the context before and after the flush. See eg the unit test:
https://github.com/elastic/go-elasticsearch/blob/0212193aad5ee67a3879b1daa24168c705d39fbd/esutil/bulk_indexer_internal_test.go#L333-L343

An example of using them for APM instrumentation is available in:

https://github.com/elastic/go-elasticsearch/blob/32c0278b8105373cee0aaac0310a41c53306f5f3/_examples/bulk/kafka/kafka.go#L106-L112

esutil/bulk_indexer.go
```diff
@@ -0,0 +1,509 @@
...
+	HasErrors bool                                 `json:"errors"`
+	Items     []map[string]BulkIndexerResponseItem `json:"items,omitempty"`
+}
+
+// BulkIndexerResponseItem represents the Elasticsearch response item.
+//
+type BulkIndexerResponseItem struct {
+	Index      string `json:"_index"`
+	DocumentID string `json:"_id"`
+	Version    int64  `json:"_version,omitempty"`
```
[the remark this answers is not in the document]
karmi: Removed in 3def616.

esutil/bulk_indexer.go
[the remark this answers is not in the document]
karmi: In the end, we went full length with callbacks, adding also `OnFlushStart()` and `OnFlushEnd()`. Many thanks for the conversation again!

esutil/bulk_indexer.go
karmi: Fixed in 93042d7.

esutil/bulk_indexer.go
karmi: Instrumentation support added in https://github.com/elastic/go-elasticsearch/commit/aa612f9daef5f04549ab8eca251cdf20eeb755f7, example available in:

https://github.com/elastic/go-elasticsearch/blob/0258f7a7f16791c3eaed36307f784efd9639b346/_examples/bulk/kafka/kafka.go#L142-L148

esutil/bulk_indexer.go
karmi: @urso, I've added a `done` channel to the indexer, which is notified in `Close()` in d294483. I've verified with https://github.com/uber-go/goleak in the corresponding test added in 9cb7b6a that it doesn't leak a goroutine. I assume, however, that since `worker.run()` is not currently able to return early, that it will block closing or leak a goroutine. Maybe that's something we can chat about.

urso: Given that we add more callbacks, how about an 'OnCancelled' callback, that allows me to capture events that have not been published yet when I shutdown the ES output. Use-case is again reconfiguration:
1. block publishing new events 
2. create new BulkIndexer
3. close old BulkdIndexer
4. forward all cancelled events to the new BulkIndexer
5. unblock publishing new events.

esutil/bulk_indexer.go
```diff
@@ -0,0 +1,545 @@
...
+			if err := w.flush(ctx); err != nil {
+				w.mu.Unlock()
+				if bi.config.OnError != nil {
+					bi.config.OnError(ctx, err)
+				}
+				continue
+			}
+		}
+		w.mu.Unlock()
+	}
```
urso: How about not forcing a flush, but have events to be marked/reported as cancelled?

esutil/bulk_indexer.go
```diff
@@ -0,0 +1,545 @@
...
+	return nil
+}
+
+// Close stops the periodic flush, closes the indexer queue channel,
+// notifies the done channel and calls flush on all writers.
+//
+func (bi *bulkIndexer) Close(ctx context.Context) error {
+	bi.ticker.Stop()
+	close(bi.queue)
+	bi.done <- true
```
urso: maybe we still should protect users from calling Close multiple times. E.g. by returning an error?

esutil/bulk_indexer.go
```diff
@@ -0,0 +1,545 @@
...
+//
+func (w *worker) flush(ctx context.Context) error {
+	if w.bi.config.OnFlushStart != nil {
+		ctx = w.bi.config.OnFlushStart(ctx)
+	}
+	defer func() {
+		if w.bi.config.OnFlushEnd != nil {
+			w.bi.config.OnFlushEnd(ctx)
+		}
+	}()
```
urso: nit; defer is on function level:

```
if w.bi.config.OnFlushEnd != nil {
  defer func() { ... }
}
```

karmi: Thanks, fixed in ef3cdce.

karmi:
Commit history cleaned up and merged as 56b3816 and aae4e81.

PR #146, opened 2020-04-26
Util: Fix for writeMeta in BulkIndexer helper

kiivihal:
This fix adds '_index' to meta when BulkIndexerItem.Index is not empty.

Currently, this value was ignored which leads to index errors when the
default index in BulkIndexerConfig.Index is empty.


kiivihal:
I have signed the Contributer Agreement.

karmi:
Hello, thanks for the patch! I've rearranged and simplified the unit test a little bit, and added a small integration test for the feature.

PR #164, opened 2020-06-25
Reinstate item.Body after it is consumed

l1redd:
This addresses the bug documented here: https://github.com/elastic/go-elasticsearch/issues/161

Issue #161, opened 2020-06-16
[BUG] BulkIndexerItem body io.Reader used before OnFailure is called


l1redd:
It is impossible to read from the `item.Body` `io.Reader` that is in the function signature of `OnFailure` in the `BulkIndexerItem` because it has already been read from in the `esutil/bulk_indexer.go` code. 

The item.Body io.Reader is read from here:
https://github.com/elastic/go-elasticsearch/blob/68f14edba6d35cbf14652c26f89f2e99b5e4cb6c/esutil/bulk_indexer.go#L415

Then this same item is used as an argument in the OnFailure function:
https://github.com/elastic/go-elasticsearch/blob/68f14edba6d35cbf14652c26f89f2e99b5e4cb6c/esutil/bulk_indexer.go#L524

So when I try to access it in the OnFailure function, it returns an empty string. For example:
```go
err = bi.Add(
	ctx,
	esutil.BulkIndexerItem{
		Action:     "index",
		Body:       bytes.NewReader(someBytes),
		DocumentID: "1",
		OnFailure: func(ctx context.Context, item esutil.BulkIndexerItem, res esutil.BulkIndexerResponseItem, err error) {

			if err != nil {
				log.Printf("error")
			} else {
				log.Printf("error in response body")
			}
			// **** Reading from item.Body returns an empty string***
			bodyBytes, err := ioutil.ReadAll(item.Body)
                        bodyString := string(bodyBytes) // bodyString is ""
                        log.Printf("Body: %s", bodyString)
		},
	},
)
```

Can I get some help on fixing this bug. I think the body needs to be saved when it's read, and then the item's body io.Reader  just needs to be reinstantiated with that body before being used as an argument in OnFailure. Happy to submit a bug fix PR if that would be faster. Thanks!

l1redd:
@karmi

karmi:
You're right, the item body is consumed when it's serialized and sent over the wire. To access it from the callback, it needs to be duplicated and reinstated indeed, in a similar manner to what happens in https://github.com/elastic/go-elasticsearch/blob/68f14edba6d35cbf14652c26f89f2e99b5e4cb6c/estransport/estransport.go#L248 or https://github.com/elastic/go-elasticsearch/blob/68f14edba6d35cbf14652c26f89f2e99b5e4cb6c/estransport/estransport.go#L264

[...]

PR #165, opened 2020-06-26
BulkIndexer - Reinstate item.Body after it is consumed

l1redd:
This addresses the bug documented here: #161
Duplicate of https://github.com/elastic/go-elasticsearch/pull/164

### DIFF
--- a/esutil/bulk_indexer.go
+++ b/esutil/bulk_indexer.go
@@ -10,6 +10,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net/http"
 	"runtime"
 	"strconv"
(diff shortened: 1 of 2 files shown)


l1redd:
I signed the CLA! Not sure when that gets reflected here.

l1redd:
The lint failures have to do with two files that I didn't touch. I'm assuming, that I shouldn't address those.

karmi:
You're right — the linter complains about a struct field in the `esapi` package, which is something which has to be addressed in the generator. I'll look into that soon.

Many thanks for the patch! I've tweaked the commit message a bit, to make it more consistent with the repo layout, and merged into the main and `7.x` branches.

karmi:
Merged via 8413c97f and 09b44b16.

l1redd:
thanks so much!

Issue #166, opened 2020-07-01
Got 'An HTTP line is larger than 4096 bytes' error when use scroll mechanism


DmSide:
Get the scroll_id by the request:
```
res, err := c.Search(
		c.Search.WithContext(background),
		c.Search.WithIndex(elasticIndex),
		c.Search.WithBody(&query),
		c.Search.WithSize(10000),
		c.Search.WithScroll(30*time.Second),
	)
```
After try to get second batch by the request:
```
res, err := c.Scroll(
			c.Scroll.WithScrollID(scrollId),
			c.Scroll.WithScroll(30*time.Second),
		)
```
and got an error  'An HTTP line is larger than 4096 bytes'

karmi:
Hi! This is a known issue. The client is 1:1 to the ES API, and therefore `c.Scroll.WithScrollID()` will pass the value as an URL parameter, which will break for large values. The solution is to pass the value in the `Body`.

karmi:
Hello! Getting back to this issue. As I was suggesting on the related pull request, the only solution here is to warn people that they should use the body, not the URL parameter. To really hit the issue, the [specification](https://github.com/elastic/elasticsearch/blob/8bdf36d18b5f013f97b8cc82efc2e076df5f2c6d/rest-api-spec/src/main/resources/rest-api-spec/api/scroll.json#L42-L45) would have to be updated to mark the URL parameter as deprecated. I'll raise this internally.

Anaethelion:
As a solution has been provided, I'm going to close this issue.
If anyone stumbles upon this, please prioritize the use of the body for long scroll ids.

PR #171, opened 2020-07-09
Scroll id in the body instead of query

sidenko-dmitrii:
This fix by the issue https://github.com/elastic/go-elasticsearch/issues/166
made by example in the documentation https://www.elastic.co/guide/en/elasticsearch/reference/6.8/search-request-scroll.html. 
The scroll should be the post method and the scroll id should be in the body.

### DIFF
--- a/esapi/api.dangling_indices.delete_dangling_index.go
+++ b/esapi/api.dangling_indices.delete_dangling_index.go
@@ -16,7 +16,7 @@ import (
 
 func newDanglingIndicesDeleteDanglingIndexFunc(t Transport) DanglingIndicesDeleteDanglingIndex {
 	return func(index_uuid string, o ...func(*DanglingIndicesDeleteDanglingIndexRequest)) (*Response, error) {
-		var r = DanglingIndicesDeleteDanglingIndexRequest{IndexUuid: index_uuid}
+		var r = DanglingIndicesDeleteDanglingIndexRequest{IndexUUID: index_uuid}
 		for _, f := range o {
 			f(&r)
 		}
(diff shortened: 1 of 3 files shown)


karmi:
Hi @sidenko-dmitrii , the entire Go API is generated from a JSON specification in the Elasticsearch repository, in this case [`scroll.json`](https://github.com/elastic/elasticsearch/blob/master/rest-api-spec/src/main/resources/rest-api-spec/api/scroll.json). The generator does support overrides, but in this particular case, it doesn't seem like we can and should do anything. The specification does mark the `scroll_id` URL part as deprecated — I'll have a look into marking it accordingly in the Go source code during generation.

Ultimately, the package should provide a helper method to facilitate using the API, similar to `esutil.BulkIndxer`.

sidenko-dmitrii:
@karmi 
The deprecated part is using the scroll_id in path and query:
https://www.elastic.co/guide/en/elasticsearch/reference/current/scroll-api.html#scroll-api-path-params
https://www.elastic.co/guide/en/elasticsearch/reference/current/scroll-api.html#scroll-api-query-params

DmSide:
@karmi Please take a look on my PR. It is okay. Do you have any comments?

karmi:
As I was saying, the entire Go API is generated, so changing the generated code doesn't really solve the issue (see the comment at the beginning of the file).

Further, the client _intentionally_ stays from any de/serialization, so calling `json.Marshal()` is not applicable.

The only fix we can do in the package is to annotate deprecated arguments, and provide guidelines. A related effort has to happen in the Elasticsearch repository, to properly deprecate and remove the arguments from the URL.

Anaethelion:
Closing this as everything has been said and `api.scroll.go` now uses the `POST` method.

cettoana:
Adding a mention of this issue to the document would be appreciated, cause the Scroll method of the Typed API is completely non-functional, and the user will only get something like

`403, failed: [security_exception], reason: action [indices:data/read/scroll] is unauthorized for user [username] ... `

PR #173, opened 2020-07-17
Automatically generate DocumentID

s0nnse:
when DocumentID is emty，it will Automatically generate the DocumentID.

Automatically generate the Document ID

POST /test_index/test_type
{
“test_content”: “my test”
}


s0nnse:
signed already

karmi:
To let Elasticsearch automatically generate the ID, please use the "Index" API, not the "Create" API — if I understand correctly what you're trying to achieve. Also, the entire `esapi` package is generated from specification, so any change would have to happen in the generator.

s0nnse:
Yes, you are right. Thank you very much

PR #176, opened 2020-08-03
Util: Allow setting version metadata in Bulk Indexer

### LINKED ISSUE #175: Bulk index util should allow specifying version and version type
The Bulk API supports document versioning metadata. These params should be exposed in the BulkIndexerItem struct.

https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html#bulk-versioning

munkyboy:
fixes #175

### DIFF
--- a/esutil/bulk_indexer.go
+++ b/esutil/bulk_indexer.go
@@ -93,6 +93,8 @@ type BulkIndexerItem struct {
 	Action          string
 	DocumentID      string
 	DocumentType    string
+	Version         int64
+	VersionType     string
 	Body            io.Reader
 	RetryOnConflict *int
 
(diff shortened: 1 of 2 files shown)


viraths:
Can we please merge this. I need this as well for handling versions.

karmi:
Hello, there's been a similar PR lately, and this needs to be addressed in a more robust way, allowing to set other supported parameters for the operation (routing, ...). I'm on vacation currently, I'll have a look at it when I'm back.

karmi:
Hello, I'm no longer maintaining the package, so I can't set a date for when this might be picked up. Ultimately, what we need here is some code generation for the bulk item struct, taking an input in JSON/YAML/etc, or other machine readable data, to keep the support in sync with the Elasticsearch Bulk API.

munkyboy:
handled in #265

PR #184, opened 2020-09-17
Fixed invalid json generation when documentid=""

phsm:
Hello, 

I've found a small bug in version 6.x BulkIndexer:
If you set DocumentType, but do not set DocumentID, then the following JSON will be generated:
```json
{"index":{"_type":"_doc",}}
```
Note the comma after "_doc". 

This small patch fixes it.


phsm:
I've signed the Contributor Agreement, 
could you please recheck?

karmi:
Hello, thanks for the patch! I've merged it into `6.x` and added a unit test to cover the case.

PR #197, opened 2020-11-04
move away from set-env in test-unit-tip workflow

Mpdreamz:
Move to new syntax: https://github.com/actions/toolkit/blob/main/docs/commands.md#environment-files


delvedor (approved):
LGTM

PR #204, opened 2020-11-29
Feature/fix actions

tty2:
Actions fail in current `master` branch. You can see it in `actions` tab or badges on main (README) page.
There are two issues that lead to this [article](https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/).

These issues are indirectly related with version of go actions. So the fix is simple: up go actions version.


sethmlarson (approved):
LGTM, thanks!

PR #206, opened 2020-12-10
Transport: remove redundant error in return from getNodeURL

tty2:
There are no  error prone cases in getNodeURL.
So the second returning parameter looks redundant.

### DIFF
--- a/estransport/discovery.go
+++ b/estransport/discovery.go
@@ -156,18 +156,14 @@ func (c *Client) getNodesInfo() ([]nodeInfo, error) {
 
 	for id, node := range nodes {
 		node.ID = id
-		u, err := c.getNodeURL(node, scheme)
-		if err != nil {
-			return out, err
-		}
-		node.URL = u
+		node.URL = c.getNodeURL(node, scheme)
 		out = append(out, node)
 	}
 
 	return out, nil
 }
 
-func (c *Client) getNodeURL(node nodeInfo, scheme string) (*url.URL, error) {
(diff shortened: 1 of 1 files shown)


karmi:
Just a small note on the thinking behind the original implementation: while it doesn't return, in fact, any error at the moment, it's easy to imagine that it might in future, as any `getXYZ()` method is prone to. So it's more or less a defensive way of coding. (When it would, later, return an error, the calling code should correctly handle it already.)

Of course, an equally valid approach is to return just the value, and when the need to return an error arises, change the method signature — since it's unexported  —, and add the error handling in the calling code. So, in other words, no problems with the patch.

(Small note about commit message again, notice how changes to the `estransport` sub-package are prefixed with `Transport: `. It makes it easier to read and discuss the Git log over time.)

tty2:
I completely agree with the point that `MethodName(params) (returningObject, error)` is a common case but if you are make it for public methods. But not for private. We can change it any time when refactor the code. Currently it's even a bit confusing (you expect something from the method signature) and pushes you to write other redundant code.

> So it's more or less a defensive way of coding.

For public method it make sense. Just in order do not change major version in future.

> (When it would, later, return an error, the calling code should correctly handle it already.)

We can easily change the method and parameters then.

> (Small note about commit message again, notice how changes to the estransport sub-package are prefixed with Transport: . It makes it easier to read and discuss the Git log over time.)

Thanks! I'll change it.

And thanks for your comment.

karmi:
Right, as I was saying above, no worries on my side with the patch!

Anaethelion:
LGTM

PR #207, opened 2020-12-11
Transport: metrics.Metrics.String remove redundant index increment in the loop

tty2:
The removed line does nothing.
Index `i` increments in `range` clause and increment inside the range clause body doesn't affect it.


MattDevy:
Hi @tty2 thank you for this contribution! This has since been fixed so I will close this PR.

PR #210, opened 2020-12-23
Only replace methodName prefix if would be empty

sethmlarson:
Some APIs like `rollup.rollup` are the same as their namespace and if we remove all occurrences of the namespace from the `methodName` then it'd result in an empty string. If we hit a situation like this only replace the namespace.

### DIFF
--- a/internal/cmd/generate/commands/genstruct/command.go
+++ b/internal/cmd/generate/commands/genstruct/command.go
@@ -239,6 +239,11 @@ type API struct {
 			name := strings.ReplaceAll(e.Name(), "Request", "")
 			if strings.HasPrefix(strings.ToLower(name), strings.ToLower(n)) {
 				methodName := strings.ReplaceAll(name, n, "")
+				// Some methods are equal to the namespace (like 'rollup.rollup')
+				// and we don't want to have an empty string here.
+				if len(methodName) == 0 {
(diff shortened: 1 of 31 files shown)


internal/cmd/generate/commands/genstruct/command.go
```diff
@@ -239,6 +239,11 @@ type API struct {
 			name := strings.ReplaceAll(e.Name(), "Request", "")
 			if strings.HasPrefix(strings.ToLower(name), strings.ToLower(n)) {
 				methodName := strings.ReplaceAll(name, n, "")
+				// Some methods are equal to the namespace (like 'rollup.rollup')
+				// and we don't want to have an empty string here.
+				if len(methodName) == 0 {
+				    methodName = strings.Replace(name, n, "", 1)
```
swallez, line 245: What if we just always used `methodName = strings.Replace(name, n, "", 1)`, i.e. always replace the first occurrence, since this is what we do anyway if the name appears several times?

sethmlarson: It's because there's always a prefix and sometimes the namespace appears within the methodName as well (ie `RollupGetRollupJob` -> `GetJob`

sethmlarson:
Closing this PR as CI failures on 8.x are being difficult. Instead focusing on 7.x.

PR #212, opened 2020-12-28
[7.x] Only replace methodName prefix if would be empty

sethmlarson:


Mpdreamz (approved):
LGTM

internal/cmd/generate/commands/gentests/skips.go
```diff
@@ -31,6 +31,8 @@ var skipFiles = []string{
 	"ml/evaluate_data_frame.yml", // Floats as map keys
 
 	"watcher/stats/10_basic.yml", // Sets "emit_stacktraces" as string ("true"), not bool
+	
+	"search.highlight/20_fvh.yml", // bad backslash
```
Mpdreamz, line 35: This particular change might need backporting to other branches?

sethmlarson:
So ML integration tests are failing in this PR for reasons I can't figure out without more knowledge on debugging Go CI, however most of the API integration tests (specifically the ones we're adding here) are running and passing so I'm going to merge this now to get a valid `7.11` branch going.

internal/cmd/generate/commands/genstruct/command.go
```diff
@@ -239,6 +239,11 @@ type API struct {
 			name := strings.ReplaceAll(e.Name(), "Request", "")
 			if strings.HasPrefix(strings.ToLower(name), strings.ToLower(n)) {
 				methodName := strings.ReplaceAll(name, n, "")
+				// Some methods are equal to the namespace (like 'rollup.rollup')
```
Mpdreamz, line 242: These changes would need a backport to other active integration branches too I assume?

internal/cmd/generate/commands/gentests/skips.go
sethmlarson: Yeah, I'll get it backported to all the relevant branches.

PR #222, opened 2021-02-09
Update generator

Anaethelion:
The generator hangs on the Rollup func being named the same as the Rollup namespace.

This fixes that and generate a new set of apis based on the updated spec.


sethmlarson:
Btw I have this commit on `7.x` and `7.11` branches: https://github.com/elastic/go-elasticsearch/commit/3011c0845d704c1680f3114cf8940cb0e408d7f1

Not sure if you want to use the same commit or replace the old code with this new method, either is fine!

Anaethelion:
If that's all the same to you I'd prefer to use the method, first step to remove code duplication.

PR #229, opened 2021-02-22
Fixed invalid xcdsearch path to xkcd in _examples/xcdsearch Dockerfile

manuelbcd:
xcdsearch example: Dockerfile path to xkcd in _examples/xcdsearch was wrong


manuelbcd:
Contributor agreement signed

Anaethelion:
Nice catch, thanks for the contribution !

PR #236, opened 2021-03-05
Implement the X-Elastic-Client-Meta HTTP header

Anaethelion:


estransport/estransport.go
```diff
@@ -194,6 +200,11 @@ func New(cfg Config) (*Client, error) {
 // Perform executes the request and returns a response or error.
 //
 func (c *Client) Perform(req *http.Request) (*http.Response, error) {
+	_, file, no, ok := runtime.Caller(-1)
```
Mpdreamz, line 203: I'm sure this was considered so asking to learn, whats the performance cost of this in Go?

Anaethelion: Good catch, that was not supposed to be committed !

esutil/bulk_indexer_internal_test.go
```diff
@@ -629,6 +630,111 @@ func TestBulkIndexer(t *testing.T) {
...
+			name string
+			args args
+			want string
+		}{
+			{
+				name: "Meta header is present in header",
+				args: args{
+					disableMetaHeader: false,
+				},
+				want: `.*,h=bp`,
```
sethmlarson, line 649: Just verifying that there won't be a space between `.*` and `,`? I don't know how Go's newline folding works by default :)

Anaethelion: There shouldn't be, but I'll modify the test to exclude this possibility!

sethmlarson (approved):
Implementation seems fine to me :+1:

PR #247, opened 2021-04-13
Add support for service token based authentication.

### LINKED ISSUE #246: Support for service tokens
With the new service tokens work that has been added in v7.13 this client also needs to support sending that header.

Just setting the `Authorization` header will cause an issue with validating an APIKey, because it will cause 2 authorization values to be set in the request.

This is required for Fleet Server and its usage of service tokens.

blakerouse:
This adds the ability to use a service token for authenticating with elasticsearch.

- Closes #246


Anaethelion:
LGTM !

PR #277, opened 2021-05-28
Add `Get` struct to `BulkIndexerResponseItem` to retrieve updated document when using `_source=true` in update request

hachh:
I want to retrieve updated document when executing update bulk request as describe here with `_source`: https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-update.html#docs-update-api-query-params

I added  `Source:        []string{"true"},` in my `BulkIndexerConfig` but `BulkIndexerResponseItem` struct was not able to retrieve updated document, so I added these fields.


Anaethelion:
Hi @hachh thanks for this contribution, can you add a test that would reflect the usage of these fields ?
I would also need you to sign the CLA to be able to go forward.

Thanks !

PR #289, opened 2021-06-29
Client: Implement product check on first call

Anaethelion:
AS titled

### DIFF
--- a/elasticsearch.go
+++ b/elasticsearch.go
@@ -19,11 +19,14 @@ package elasticsearch
 
 import (
 	"encoding/base64"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"net/http"
 	"net/url"
 	"os"
+	"regexp"
+	"strconv"
 	"strings"
 	"time"
 
@@ -32,8 +35,21 @@ import (
 	"github.com/elastic/go-elasticsearch/v7/internal/version"
 )
 
+var (
+	reVersion          *regexp.Regexp
+	useHeaderCheckOnly bool
+	productChecked     bool
+)
+
+func init() {
+	versionPattern := `(\d+).(\d+).(\d+)`
+	reVersion = regexp.MustCompile(versionPattern)
+}
+
 const (
 	defaultURL = "http://localhost:9200"
+	tagline    = "You Know, for Search"
+    unknownProduct = "the client noticed that the server is not Elasticsearch and we do not support this unknown product"
 )
 
 // Version returns the package version as a string.
@@ -69,7 +85,8 @@ type Config struct {
 	EnableMetrics     bool // Enable the metrics collection.
 	EnableDebugLogger bool // Enable the debug logging.
 
-	DisableMetaHeader bool // Disable the additional "X-Elastic-Client-Meta" HTTP header.
+	DisableMetaHeader  bool // Disable the additional "X-Elastic-Client-Meta" HTTP header.
+	UseHeaderCheckOnly bool
 
 	RetryBackoff func(attempt int) time.Duration // Optional backoff duration. Default: nil.
 
@@ -86,6 +103,16 @@ type Config struct {
 type Client struct {
 	*esapi.API // Embeds the API methods
 	Transport  estransport.Interface
+}
+
+type EsVersion struct {
+	Number                           string    `json:"number"`
+	BuildFlavor                      string    `json:"build_flavor"`
+}
+
+type Info struct {
+	Version     EsVersion `json:"version"`
+	Tagline     string    `json:"tagline"`
 }
 
 // NewDefaultClient creates a new client with default options.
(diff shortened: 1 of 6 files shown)


elasticsearch.go
```diff
@@ -19,21 +19,38 @@ package elasticsearch
...
+
+func init() {
+	versionPattern := `(\d+).(\d+).(\d+)`
+	reVersion = regexp.MustCompile(versionPattern)
+}
+
 const (
 	defaultURL = "http://localhost:9200"
+	tagline    = "You Know, for Search"
+    unknownProduct = "the client noticed that the server is not Elasticsearch and we do not support this unknown product"
```
sethmlarson, line 53: Are error messages typically lowercased in Go?

elasticsearch.go
```diff
@@ -19,21 +19,38 @@ package elasticsearch
...
 )
 
+var (
+	reVersion          *regexp.Regexp
+	useHeaderCheckOnly bool
+	productChecked     bool
+)
+
+func init() {
+	versionPattern := `(\d+).(\d+).(\d+)`
```
sethmlarson, line 46: Do we need a `^` to start this pattern? Also `.` should be escaped so it's not used as any character.

Does Go do special things with unicode when using `\d` like Python does? If so `[0-9]` may be preferable.

elasticsearch.go
```diff
@@ -69,7 +86,8 @@ type Config struct {
 	EnableMetrics     bool // Enable the metrics collection.
 	EnableDebugLogger bool // Enable the debug logging.
 
-	DisableMetaHeader bool // Disable the additional "X-Elastic-Client-Meta" HTTP header.
+	DisableMetaHeader  bool // Disable the additional "X-Elastic-Client-Meta" HTTP header.
+	UseHeaderCheckOnly bool
```
sethmlarson, line 90: This option is pretty opaque based on its name, just throwing out `CheckOnlyElasticProductHeader` for inspiration on a potential better name?

elasticsearch.go
```diff
@@ -189,12 +218,107 @@ func NewClient(cfg Config) (*Client, error) {
...
+	return nil
+}
+
+func genuineCheckInfo(info info) error {
+	major, minor, _, err := ParseElasticsearchVersion(info.Version.Number)
+	if err != nil {
+		return err
+	}
+
+	if major < 6 {
```
sethmlarson, line 242: This function might do well with some prose comments on what is checked for which version ranges?

elasticsearch.go
```diff
@@ -189,12 +218,107 @@ func NewClient(cfg Config) (*Client, error) {
...
+	minor, _ := strconv.ParseInt(matches[2], 10, 0)
+	patch, _ := strconv.ParseInt(matches[3], 10, 0)
+
+	return major, minor, patch, nil
 }
 
 // Perform delegates to Transport to execute a request and return a response.
 //
 func (c *Client) Perform(req *http.Request) (*http.Response, error) {
+	if !productChecked {
```
sethmlarson, line 280: Is `productChecked` supposed to be a global or an attribute of the `Client`?

elasticsearch.go
```diff
@@ -189,12 +218,107 @@ func NewClient(cfg Config) (*Client, error) {
...
+
+			switch res.StatusCode {
+			case http.StatusForbidden:
+			case http.StatusUnauthorized:
+				break
+			default:
+				err = genuineCheckHeader(res.Header)
+
+				if err != nil {
+					if !useHeaderCheckOnly && info.Version.Number != "" {
```
sethmlarson, line 308: Do we need to special case the empty string for `info.Version.Number` if it'll be handled within `genuineCheckInfo()`?

sethmlarson (changes requested):
Some initial things I found, let me know if some of these are Go-related and I don't understand, very possible!

elasticsearch_internal_test.go
```diff
@@ -22,15 +22,49 @@ package elasticsearch
...
 )
 
+var called bool
+
+type mockTransp struct{
+	RoundTripFunc func(*http.Request) (*http.Response, error)
+}
+
+var defaultRoundTripFunc = func(req *http.Request) (*http.Response, error) {
+	response := &http.Response{Header: http.Header{"X-Elastic-Product": []string{"X-Elastic-Product"}}}
```
sethmlarson, line 42: Should this be `X-Elastic-Product: Elasticsearch`?

elasticsearch.go
Anaethelion: Yep, they shouldn't end with punctuation either, see https://github.com/golang/go/wiki/CodeReviewComments#error-strings

elasticsearch.go
Anaethelion: It's a global variable declared at the top.

elasticsearch.go
Anaethelion: In Go if the structure is not initialized it's field have default type values, in this case `string` is just an empty string.

sethmlarson (approved):
LGTM, wish Python had a `sync.Once` construction in the standard library!

PR #303, opened 2021-07-09
Product check fixes

axw:
- perform cheaper bool checks before synchronisation methods
- don't issue original request if product check fails
- check original request error before response header check (avoiding panic)
- replace `sync.Once` approach so transient network errors are not cached


Anaethelion:
LGTM!

PR #309, opened 2021-07-21
Product check: Improve retry mechanism 

Anaethelion:
This ensures the product check is not flagged wrongfully as successful if the answer has not been properly validated.
The client will retry product check until success or error.

### DIFF
--- a/elasticsearch.go
+++ b/elasticsearch.go
@@ -50,7 +50,7 @@ const (
 	defaultURL         = "http://localhost:9200"
 	tagline            = "You Know, for Search"
 	unknownProduct     = "the client noticed that the server is not Elasticsearch and we do not support this unknown product"
-	unsupportedProduct = "The client noticed that the server is not a supported distribution of Elasticsearch"
+	unsupportedProduct = "the client noticed that the server is not a supported distribution of Elasticsearch"
 )
 
 // Version returns the package version as a string.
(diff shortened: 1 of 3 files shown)


axw:
> This ensures the product check is not flagged wrongfully as successful if the answer has not been properly validated.

@Anaethelion can you describe how this was happening before? This change seems complicated to me, and I don't understand why it's necessary.

elasticsearch.go
```diff
@@ -297,9 +300,16 @@ func (c *Client) Perform(req *http.Request) (*http.Response, error) {
 
 	// ResponseCheck path continues, we run the header check on the first answer from ES.
 	if err == nil {
-		checkHeader := func() error { return genuineCheckHeader(res.Header) }
+		checkHeader := func() error {
+			if res != nil {
+				return genuineCheckHeader(res.Header)
+			}
+			return nil
+		}
```
axw, line 308: As we have already checked `err == nil` above, res must be non-nil according to [net/http](https://pkg.go.dev/net/http#Client.Do):

> If the returned error is nil, the Response will contain a non-nil Body which the user is expected to close. If the Body is not both read to EOF and closed, the Client's underlying RoundTripper (typically Transport) may not be able to re-use a persistent TCP connection to the server for a subsequent "keep-alive" request.

axw (approved):

elasticsearch.go
Anaethelion: Nice catch, thanks for this! 👍

PR #315, opened 2021-07-28
[BulkIndexer] Use ReadSeeker for BulkIndexerItem body

Anaethelion:
Allows to reuse the existing reader in callbacks without allocating a new buffer.
Should reduce memory footprint on large scale bulk indexers.

### DIFF
--- a/esutil/bulk_indexer.go
+++ b/esutil/bulk_indexer.go
@@ -23,7 +23,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"net/http"
 	"runtime"
 	"strconv"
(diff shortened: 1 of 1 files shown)


qu0laz:
This one change has made a significant impact on memory footprint. For me specifically, the bulk indexer would consume the 11gb of memory available on the machine and force a hard shutdown. Typically would occur after ~1.5m docs were inserted to elastic.
Screenshot from 2021-09-03 12-17-41
After implementing this one fix the memory has been greatly reduced.
Screenshot from 2021-09-03 12-18-36

I'm not as well skilled in go as others here, so I can't be 100% sure that nothing else broke because of the change or if there is still a small memory leak. However I have pushed through 100m docs in the last 24 hours and it has not yet run out of memory.

PR #321, opened 2021-08-03
Thread request context through to product check

axw:
This is important for ensuring request cancellation, as well as for tracing product check requests.


Anaethelion:
LGTM!

PR #348, opened 2021-10-19
Style: format code

Loyalsoldier:
Run following commands in the project root:

- `go fmt ./...`
- `gofmt -s -l -e -w $(find . -type f -name "*.go")`
- `go install -v github.com/daixiang0/gci@latest` and `find . -type f -name "*.go" | xargs gci -w`

### DIFF
--- a/_benchmarks/benchmarks/actions/001_ping.go
+++ b/_benchmarks/benchmarks/actions/001_ping.go
@@ -21,10 +21,9 @@ import (
 	"io"
 	"io/ioutil"
 
-	"github.com/elastic/go-elasticsearch/v8/esapi"
-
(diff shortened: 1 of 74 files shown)


Loyalsoldier:
> Hi @Loyalsoldier
> 
> Thank you for this PR, however I'd prefer to stick to the standard tooling regarding format rules. Can you remove the commit using the `gci` ?

Actually what `gci` does here is just to make the package import order deterministic (remove newlines introduced intentionally or unintentionally by different developers).

The official tooling keeps all newlines between standard packages and all third-party packages, which will be messy at last if many developers edited the same file by different text editors/IDEs with different Go plugins.

If you remove all newlines between standard packages and third-party packages manually, and format them again, the `gci` and official tooling perform the same output.

PR #397, opened 2022-01-11
Fix "concurrent map iteration and map write" while read estransport.M…

mainliufeng:
"fatal error: concurrent map iteration and map write" while other goroutine reads estransport.Metrics.Responses

```
func TestTransportPerformAndReadMetricsResponses(t *testing.T) {
	t.Run("Executes", func(t *testing.T) {
		u, _ := url.Parse("https://foo.com/bar")
		tp, _ := New(Config{
			EnableMetrics: true,
			URLs:          []*url.URL{u},
			Transport: &mockTransp{
				RoundTripFunc: func(req *http.Request) (*http.Response, error) { return &http.Response{Status: "MOCK"}, nil },
			}})

		go func() {
			for {
				metrics, _ := tp.Metrics()
				t.Logf("len(responses): %v", len(metrics.Responses))
				for code, num := range metrics.Responses {
					t.Logf("%v(%v)", code, num)
				}
			}
		}()

		for i := 0; i < 100000; i++ {
			req, _ := http.NewRequest("GET", "/abc", nil)
			res, err := tp.Perform(req)
			if err != nil {
				t.Fatalf("Unexpected error: %s", err)
			}

			if res.Status != "MOCK" {
				t.Errorf("Unexpected response: %+v", res)
			}
		}
	})
}
```

### DIFF
--- a/estransport/estransport_internal_test.go
+++ b/estransport/estransport_internal_test.go
@@ -1109,3 +1109,37 @@ func TestRequestCompression(t *testing.T) {
 		})
 	}
 }
+
+func TestTransportPerformAndReadMetricsResponses(t *testing.T) {
+	t.Run("Read Metrics.Responses", func(t *testing.T) {
+		u, _ := url.Parse("https://foo.com/bar")
+		tp, _ := New(Config{
+			EnableMetrics: true,
+			URLs:          []*url.URL{u},
+			Transport: &mockTransp{
+				RoundTripFunc: func(req *http.Request) (*http.Response, error) { return &http.Response{Status: "MOCK"}, nil },
+			}})
+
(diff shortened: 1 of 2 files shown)


Anaethelion:
Hi @mainliufeng 

Thanks for raising this and the patch that goes with it.
Copying the map to the new structure seems fine to me, do you think you can integrate the test you highlighted in the description to the code so we can prevent regression in the future ?

mainliufeng:
@Anaethelion I add a unit test to estransport/estransport_internal_test.go

Anaethelion:
Jenkins test this please

Anaethelion:
Failing CI is already fixed in 7.17 branch.
LGTM

PR #475, opened 2022-04-13
fix concurrent map write error caused when assigned header

chzhuo:
image

line 533 will panic when assigned config.Header


Anaethelion:
Hi @chzhuo 

Could you add a test that would highlight this ?
I would also need you to sign the CLA before going further with this.

Thank you!

Anaethelion:
Jenkins test this please

PR #493, opened 2022-05-12
replace readme string concat with json marshal

epelc:
Just took a look after being a long time user of oliver/elastic-go and noticed a quite jarring promotion of string concatenation in the example for indexing a document.

I don't think a database driver should be promoting usage of string concatenation. Someone is going to copy this into production and end up with an injection vulnerability or request smuggling issue.

https://github.com/elastic/go-elasticsearch#usage

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,6 +196,7 @@ import (
   "strconv"
   "strings"
   "sync"
+  "bytes"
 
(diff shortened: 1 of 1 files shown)


epelc:
signed the cla

PR #516, opened 2022-07-28
Use pkg.go.dev instead of godoc.org on overview page

AndersonQ:
The [README.md](https://github.com/elastic/go-elasticsearch#readme) is already updated, but [Elastic Docs](https://www.elastic.co/guide/en/elasticsearch/client/go-api/master/overview.html) isn't.


Anaethelion:
Nice catch, thank you!

PR #536, opened 2022-10-05
API: only set default content-type when not set

### LINKED ISSUE #132: Do we support CBOR format communicating
hi guys,

I just noticed https://github.com/elastic/go-elasticsearch/blob/v6.8.5/esapi/api.search.go#L316, seems all the data format we could support is json.

and from the official document https://www.elastic.co/guide/en/elasticsearch/reference/6.8/common-options.html#_content_type_requirements. I got a little confused, will we support the data format other than json in this library.

pzl:
This updates the API methods to only set the default content type (json) _when_ the content-type header has not already been manually specified.

This allows a client to manually specify if they are using a different request body format (e.g. yaml, cbor, smile).

This change is all that was needed to successfully send CBOR from Fleet Server (https://github.com/elastic/fleet-server/pull/1902) into elasticsearch

This (i think) closes #132

### DIFF
--- a/elasticsearch_internal_test.go
+++ b/elasticsearch_internal_test.go
@@ -27,6 +27,7 @@ import (
 	"crypto/x509"
 	"encoding/base64"
 	"errors"
+	"github.com/elastic/go-elasticsearch/v8/esapi"
 	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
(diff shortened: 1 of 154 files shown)


pzl:
FYI the failures here are from some out-of-date API stuff that gets pulled down from `make gen-api`.

There were a few new files and a couple of parameter updates (and several things moved from experimental to beta). I excluded those things from this PR since they seemed like in-tree updates and chose to kept this PR thematically one-note.

CI didn't like that. Happy to just make an update PR and then rebase this, if that's ok, @Anaethelion

Anaethelion:
@pzl could you just remove the API generation from the PR and keep only the part within the generator ?
I'll take care of updating the API for main and 8.5 once this is merged!

Anaethelion (approved):
LGTM!

Anaethelion:
Tests are failing because API is not in sync anymore with tests, merging and backporting as-is. API will be updated in branch depending on their respective versions.

PR #575, opened 2022-12-22
Reduce CPU usage

rockdaboot:
My client application writes a lot of relatively small documents (few hundred bytes each) into ES. Without request compression, the bulk indexer is responsible for >90% of the CPU spent by the client. With request compression it is less, as ~20% of the CPU time is now spent inside the gzip compression. It still seemed worth improving the the bulk indexer code...

The application writes ~3 million documents to a remote ES cluster.

### Timing based on `main` (commit 30b59239b0863522ac7edf7cbb8372a6441e8439)
```
real    4m43.854s
user    5m41.925s
sys     0m23.021s
(pprof) Duration: 283.71s, Total samples = 360.56s (127.09%)
```

### Timing based on the first commit (Compute length using Seek())
```
real    4m44.704s
user    5m4.591s
sys     0m20.914s
(pprof) Duration: 284.55s, Total samples = 322.17s (113.22%)
```

### Timing based on both commits
```
real    4m41.193s
user    4m42.874s
sys     0m19.013s
(pprof) Duration: 281.10s, Total samples = 298.95s (106.35%)
```

Both commits reduce the number of heap allocations in the hot path.
The performance improvements are gained mostly by less (expensive) calls to gcmalloc()+memmove and consequently by a reduced GC pressure.

The real time stays relatively constant, mostly due to network throughput and latency plus ES response times being constant.

### DIFF
--- a/esutil/bulk_indexer.go
+++ b/esutil/bulk_indexer.go
@@ -21,7 +21,6 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io"
 	"net/http"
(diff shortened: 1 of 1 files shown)


Anaethelion (approved):
LGTM
Thanks for taking the time to improve these!

PR #594, opened 2023-01-26
Add Config.CompressRequestBodyLevel

rockdaboot:
**Disclaimer**: This PR depends on https://github.com/elastic/elastic-transport-go/pull/12 to be merged.

**Story / Background**
With our large number of documents to be created, enabling the request compression
gives us ~15-20x less payload to be transferred from client to ES which translates
into ~4-5x faster wall time for the whole operation.
Of course, the latter is heavily dependent on the network speed between client and ES.

Example timings from `time` (somewhat flaky)
```
real    0m5.496s
user    0m0.903s
sys     0m0.150s
```

With compression enabled, the consumed user CPU time goes up.

Go pprof (sort=cum) shows 970ms
```
flat  flat%   sum%        cum   cum%
20ms  2.06%  2.06%      460ms 47.42%  github.com/elastic/go-elasticsearch/v8/esutil.(*worker).flush
   0     0%  2.06%      240ms 24.74%  compress/flate.(*Writer).Write (inline)
```
The bulk indexer flush() function is the most heavy function (cumulated CPU).
The biggest part here is the gzip compression.

**Test results with faster compression**
Using a compression level of `gzip.BestSpeed` instead of `gzip.DefaultCompression` only slightly increases
the payload to be transferred. But the CPU usage of the compress function goes down measureably.

Example timings from `time` (somewhat flaky)
```
real    0m4.650s
user    0m0.766s
sys     0m0.171s
```

Go pprof (sort=cum) shows 870ms total
```
flat  flat%   sum%        cum   cum%
   0     0%     0%      360ms 41.38%  github.com/elastic/go-elasticsearch/v8/esutil.(*worker).flush
   0     0%  1.15%      120ms 13.79%  compress/flate.(*Writer).Write (inline)
```

Be aware that these numbers are flaky and just give a rough direction.

**Suggested Improvement**
I'd like to suggest a new configuration option to let the user configure the compression level.

`Config.CompressRequestBodyLevel` is of type `int` to be in sync with the predefined gzip levels.
To stay compatible with with the existing behavior, the default translates into `gzip.DefaultCompression`.

### DIFF
--- a/elasticsearch.go
+++ b/elasticsearch.go
@@ -89,6 +89,7 @@ type Config struct {
 	RetryOnError  func(*http.Request, error) bool // Optional function allowing to indicate which error should be retried. Default: nil.
 
 	CompressRequestBody bool // Default: false.
+	CompressRequestBodyLevel int // Default: gzip.DefaultCompression.
 
 	DiscoverNodesOnStart  bool          // Discover nodes when initializing the client. Default: false.
 	DiscoverNodesInterval time.Duration // Discover nodes periodically. Default: disabled.
@@ -275,6 +276,7 @@ func newTransport(cfg Config) (*elastictransport.Client, error) {
 		RetryBackoff:  cfg.RetryBackoff,
 
 		CompressRequestBody: cfg.CompressRequestBody,
+		CompressRequestBodyLevel: cfg.CompressRequestBodyLevel,
 
 		EnableMetrics:     cfg.EnableMetrics,
 		EnableDebugLogger: cfg.EnableDebugLogger,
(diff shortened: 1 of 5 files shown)


rockdaboot:
@Anaethelion Can you help me figure out what that failed "client-ci" check is and how to fix it ?

Anaethelion:
> @Anaethelion Can you help me figure out what that failed "client-ci" check is and how to fix it ?

I've had a look and its not related to your changes, we can proceed as is!

Anaethelion (approved):
LGTM!

PR #595, opened 2023-02-06
Enable bulk indexer workers to do concurrent flushing

iskendermert:
When bulk indexer workers are enabled; items are processed concurrently but network transfer operations are done sequentially. This enables concurrent network transfers for bulk indexer workers.


rockdaboot:
Thanks for your contribution !

I think PR #624 makes this PR obsolete (a side effect of some restructuring). The workers are running in parallel and `Close()` signals the workers to immediately flush. Flushing is done in parallel then.

PR #600, opened 2023-02-18
Mention `Response.IsError()` also handles 3xx in example

JohnTitor:
This adds `3xx` to the `Response.IsError()` example comment as  [it catches `>299` status codes](https://github.com/elastic/go-elasticsearch/blob/11608f98966a036fb555a51e7a08955f4d042c28/esapi/esapi.response.go#L89-L93).


Anaethelion:
Agreed, thank you!

PR #602, opened 2023-02-20
Resolve a deadlock in go-elasticsearch

kuaikuai:
Deadlock in go-elasticsearch/estransport/connection.go
In func (cp *statusConnectionPool) OnSuccess(c *Connection), the lock order ( c.lock->cp.clock)  is different from other functions.


kuaikuai:
I have signed a Contributor Agreement.

Issue #274, opened 2021-05-21
Deadlock in go-elasticsearch/estransport/connection.go ？！


bjkxt:
I have some golang programs using go-elasticsearch to access my ES cluster.
I found that one of them can't work any more so I use the pprof by http://x.x.x.x/debug/pprof/goroutine?debug=1 to get the goroutines' status.
It seems that there was a deadlock between function OnSuccess() (in file estransport/connection.go:138)  and function scheduleResurrect() (in file estransport/connection.go:259).
What can I do to prevent this situation? 
Thanks a lot for your advice.

------------------------------------------
some of the pprof's returns are below：

63695 @ 0x439390 0x44a450 0x44a43b 0x44a1b7 0x48380c 0xa9c022 0xa9c02b 0x8cbb55 0x9e8742 0x117a268 0x12921a2 0x1294b47 0x469661
\#	0x44a1b6	sync.runtime_SemacquireMutex+0x46						/data/go/src/runtime/sema.go:71
\#	0x48380b	sync.(*Mutex).lockSlow+0xfb							/data/go/src/sync/mutex.go:138
\#	0xa9c021	sync.(*Mutex).Lock+0xaf1							/data/go/src/sync/mutex.go:81
\#	0xa9c02a	github.com/elastic/go-elasticsearch/v6/estransport.(*Client).Perform+0xafa	/data/GoProject/pkg/mod/github.com/elastic/go-elasticsearch/v6@v6.8.10/estransport/estransport.go:234
\#	0x8cbb54	github.com/elastic/go-elasticsearch/v6/esapi.SearchRequest.Do+0x1244		/data/GoProject/pkg/mod/github.com/elastic/go-elasticsearch/v6@v6.8.10/esapi/api.search.go:335
\#	0x9e8741	github.com/elastic/go-elasticsearch/v6/esapi.newSearchFunc.func1+0x141		/data/GoProject/pkg/mod/github.com/elastic/go-elasticsearch/v6@v6.8.10/esapi/api.search.go:25

31164 @ 0x439390 0x449513 0x7786e5 0x773feb 0x75a785 0xa9b8b5 0x8cbb55 0x9e8742 0x117a268 0x12921a2 0x1294b47 0x469661
\#	0x7786e4	net/http.(*Transport).getConn+0x584						/data/go/src/net/http/transport.go:1350
\#	0x773fea	net/http.(*Transport).roundTrip+0x76a						/data/go/src/net/http/transport.go:569
[...]

PR #603, opened 2023-02-20
Resolve a deadlock

kuaikuai:


Anaethelion:
Hi @kuaikuai 

Thank you, could you add a test and a brief description of the context this would be happening ?

kuaikuai:
https://github.com/elastic/go-elasticsearch/issues/274

kuaikuai:
the test code as the following:
package main
import (
        "os"
        "time"
        "fmt"
        "net/http"
	"net/http/pprof"
        "strings"
        "bytes"
        "sync/atomic"
	"github.com/elastic/go-elasticsearch/v7"
)

type CountingTransport struct {
    count uint64
}
func (t *CountingTransport) RoundTrip(req *http.Request) (*http.Response, error) {
    var buf bytes.Buffer
    atomic.AddUint64(&t.count, 1)

    if t.count % 500 != 0 {
        return nil, fmt.Errorf("some error")
    }
    req.Header.Set("Accept", "application/yaml")
    req.Header.Set("X-Request-ID", "foo-123")

    res, err := http.DefaultTransport.RoundTrip(req)

    buf.WriteString(strings.Repeat("-", 80) + "\n")
    fmt.Fprintf(&buf, "%s %s", req.Method, req.URL.String())

    if err != nil {
        fmt.Fprintf(&buf, "ERROR: %s\n", err)
    } else {
        fmt.Fprintf(&buf, "[%s] %s\n", res.Status, res.Header.Get("Content-Type"))
    }
    buf.WriteTo(os.Stdout)
    return res, err
}

func main() {
    tp := new(CountingTransport)
    hosts := []string{"http://x.x.x.x:9200","http://y.y.y.y:9200"}
    es, err := elasticsearch.NewClient(elasticsearch.Config{Addresses: hosts, Transport: tp, Username:"elastic", Password:"Mam123456*"})
    if err != nil {
        fmt.Printf("err: %v\n", err)
        return
    }
    for i := 0; i < 5; i++ {
        go func() {
            for {
                es.Info()
                time.Sleep(time.Millisecond * 10)
            }
        }()
    }

   r := http.NewServeMux()
   r.HandleFunc("/debug/pprof/", pprof.Index)
    http.ListenAndServe("0.0.0.0:6060", r)
}
------------------------------------------------
in addition to speed up the testing, I change estransport/connection.go:
from:
var (
	defaultResurrectTimeoutInitial      = 60 * time.Second
	defaultResurrectTimeoutFactorCutoff = 5
)
to
var (
	defaultResurrectTimeoutInitial      = 30 * time.Second
	defaultResurrectTimeoutFactorCutoff = 3
)
I don't known how long the test have ran. when I check it again, use curl command: 
curl http://127.0.0.1:6060/debug/pprof/goroutine?debug=1

1 @ 0x43c1c5 0x44db45 0x44db2e 0x46e3a7 0x47b465 0x8d22bd 0x8d223d 0x8d65b0 0x75bf48 0x89c2cd 0x8db163 0x472081
#	0x46e3a6	sync.runtime_SemacquireMutex+0x46								/home/yxf/go1.16.2/src/runtime/sema.go:71
#	0x47b464	sync.(*Mutex).lockSlow+0x104									/home/yxf/go1.16.2/src/sync/mutex.go:138
#	0x8d22bc	sync.(*Mutex).Lock+0x11c									/home/yxf/go1.16.2/src/sync/mutex.go:81
#	0x8d223c	github.com/elastic/go-elasticsearch/v7/estransport.(*statusConnectionPool).OnSuccess+0x9c	/home/yxf/compile/test/src/github.com/elastic/go-elasticsearch/v7/estransport/connection.go:138
#	0x8d65af	github.com/elastic/go-elasticsearch/v7/estransport.(*Client).Perform+0x88f			/home/yxf/compile/test/src/github.com/elastic/go-elasticsearch/v7/estransport/estransport.go:295
#	0x75bf47	github.com/elastic/go-elasticsearch/v7/esapi.InfoRequest.Do+0x487				/home/yxf/compile/test/src/github.com/elastic/go-elasticsearch/v7/esapi/api.info.go:107
#	0x89c2cc	github.com/elastic/go-elasticsearch/v7/esapi.newInfoFunc.func1+0x12c				/home/yxf/compile/test/src/github.com/elastic/go-elasticsearch/v7/esapi/api.info.go:21
#	0x8db162	main.main.func1+0x42										/home/yxf/compile/test/src/fs/main.go:57

1 @ 0x43c1c5 0x44db45 0x44db2e 0x46e3a7 0x47b465 0x8d8c5e 0x8d8b14 0x472081
#	0x46e3a6	sync.runtime_SemacquireMutex+0x46									/home/yxf/go1.16.2/src/runtime/sema.go:71
#	0x47b464	sync.(*Mutex).lockSlow+0x104										/home/yxf/go1.16.2/src/sync/mutex.go:138
#	0x8d8c5d	sync.(*Mutex).Lock+0x1bd										/home/yxf/go1.16.2/src/sync/mutex.go:81
#	0x8d8b13	github.com/elastic/go-elasticsearch/v7/estransport.(*statusConnectionPool).scheduleResurrect.func1+0x73	/home/yxf/compile/test/src/github.com/elastic/go-elasticsearch/v7/estransport/connection.go:262

kuaikuai:
I made a mistake with my first commit. In the second commit, I only changed the lock method. Can you help me review it?
@Anaethelion

Anaethelion:
I've taken some time to dig through this and I agree with @thomas-long-f3 comment.

Properly locking the connection pool should be enough to prevent the deadlock, I've pushed a change to your branch directly to reflect that.

@kuaikuai would you take some time to validate my version on your side ?

kuaikuai:
> I've taken some time to dig through this and I agree with @thomas-long-f3 comment.
> 
> Properly locking the connection pool should be enough to prevent the deadlock, I've pushed a change to your branch directly to reflect that.
> 
> @kuaikuai would you take some time to validate my version on your side ?

Thanks for your help. I have validated your version and it looks good to me.

kuaikuai:
@thomas-long-f3, the change is OK.

kuaikuai:
Thank you for you help @Anaethelion, I think this code has a hidden bug maybe.
```
func (cp *statusConnectionPool) OnSuccess(c *Connection) error {
    // Short-circuit for live connection
    if !c.IsDead {
        return nil
    }

    cp.Lock()
    defer cp.Unlock()

    c.Lock()
    defer c.Unlock()
    /*
       In certain scenarios, another OnSuccess(c) might have occurred before this pass.
       As a result, c.IsDead is now false, and c is already in cp.live. 
       Then, cp.resurrect will add c to cp.live twice, causing the connection pool to become corrupted.
    */
    c.markAsHealthy()
    return cp.resurrect(c, true)
}
```

kuaikuai:
@Anaethelion
My 2nd commit is a little complex, But the impact of its complexity is minor.
If you want the code to be simpler, I agree with @thomas-long-f3 comment too.

Anaethelion:
Acknowledged that removing this lock require some more refactoring, I'd prefer @kuaikuai version.
I'll update the PR accordingly.

kuaikuai:
I still believe that the commited code has a bug.
Why don't we check whether c.IsDead is false? 
If I am wrong, please tell me why.
```
func (cp *statusConnectionPool) OnSuccess(c *Connection) error {
	c.Lock()
	// Short-circuit for live connection
	if !c.IsDead {
		c.Unlock()
		return nil
	}
        c.Unlock()

        // At that moment between c.Unlock() and cp.Lock(), if another OnSuccess(c) is executed, it could modify cp.

	cp.Lock()
	defer cp.Unlock()
	c.Lock()
	defer c.Unlock()
	// here, why we don't check c.IsDead?
	// If the previous OnSuccess(c) has already executed cp.resurrect, now this function should not call cp.resurrect(c, true).
	c.markAsHealthy()
	return cp.resurrect(c, true)
}
```

Anaethelion:
I don't see that happening while `OnSuccess` is also called under a lock at the client level.

kuaikuai:
> I don't see that happening while `OnSuccess` is also called under a lock at the client level.
@Anaethelion
Okay, I understand now. Thank you for your help!

kuaikuai:
```
```

PR #615, opened 2023-03-06
Notify items if an error occurs in bulk indexer

### LINKED ISSUE #179: BulkIndexer do not provide info to BulkIndexerItem on global worker.flush error
Hello,

I started to use `esutil.BulkIndexer` recently (with elasticsearch v6) and got an error by not providing a type on an index action. I expected to obtain the info thanks to the `OnFailure` callback but got nothing. I now understand that `OnFailure` does not report errors that occur with the whole `worker.flush()` and that only the `BulkIndexer.OnError()` callback does so.  
I did not find any easy workaround helping me to detect when the `worker.flush()` in charge of my `BulkIndexerItem` fails.

Do you have a workaround for this situation or does it requires an evolution of the BulkIndexer?  
If it needs a PR, what do you recommend me to do?
* Calling `OnFailure` for each item when an error occurs
* Adding another callback (for example `OnFlushError`) to the items, dedicated to this situation
* Something else ...

Regards

andersonlribeiro:
Propagate errors on items. Fix #179

### DIFF
--- a/esutil/bulk_indexer.go
+++ b/esutil/bulk_indexer.go
@@ -573,7 +573,8 @@ func (w *worker) flush(ctx context.Context) error {
 		if w.bi.config.OnError != nil {
 			w.bi.config.OnError(ctx, fmt.Errorf("flush: %s", err))
 		}
-		return fmt.Errorf("flush: %s", err)
+		w.notifyItemsOnError(ctx, err)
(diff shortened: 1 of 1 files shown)


MattDevy:
Hi all, I've added some tests to cover the new behaviour, with a slight refactor to make the code more DRY and a minor bug fix. This will be sorted soon, thank you for your patience and for your contribution!

PR #623, opened 2023-03-17
Do not drop item after flush() failed

rockdaboot:
Bulk Indexer: In run() we drop the current item when the flush() failed.
This PR fixes it.


Anaethelion (approved):
Nice catch!
LGTM

PR #624, opened 2023-03-17
[Bulk Indexer] Use one ticker per worker

rockdaboot:
The ticker is currently not reset after flush. It flushes the data even if it was flushed just right before due to reaching the maximal buffer size. This is inefficient and leads to ineffective (partial) flushes.

This PR uses a ticker per worker, that can be reset with each flush. This avoids the mentioned premature flushes. A nice side-effect is that I could remove the worker mutex because now the item handling is serialized in each worker by just reading the item channel.


esutil/bulk_indexer.go
```diff
@@ -425,73 +379,74 @@ func (w *worker) run() {
 		if w.bi.config.DebugLogger != nil {
 			w.bi.config.DebugLogger.Printf("[worker-%03d] Started\n", w.id)
 		}
-		defer w.bi.wg.Done()
+		defer func() {
+			w.ticker.Stop()
+			_ = w.flush(ctx)
```
Anaethelion, line 384: What's the reasoning for no propagating the error from flush ?

esutil/bulk_indexer.go
```diff
@@ -425,73 +379,74 @@ func (w *worker) run() {
...
-		for item := range w.ch {
-			w.mu.Lock()
+		for {
+			select {
+			case <-w.ticker.C:
+				if w.bi.config.DebugLogger != nil {
+					w.bi.config.DebugLogger.Printf("[worker-%03d] Auto-flushing after %s\n",
+						w.id, w.bi.config.FlushInterval)
+				}
+				_ = w.flush(ctx)
```
Anaethelion, line 395: Same question as line 384.

esutil/bulk_indexer.go
```diff
@@ -425,73 +379,74 @@ func (w *worker) run() {
...
-					if w.bi.config.OnError != nil {
-						w.bi.config.OnError(ctx, err)
+				w.items = append(w.items, item)
+				// Should the item payload exceed the configured FlushBytes flush happens instantly.
+				if oversizePayload {
+					if w.bi.config.DebugLogger != nil {
+						w.bi.config.DebugLogger.Printf("[worker-%03d] Oversize Payload in item [%s:%s]\n", w.id, item.Action, item.DocumentID)
 					}
-					continue
+					_ = w.flush(ctx)
```
Anaethelion, line 434: Third error for flush, marked for reference.

Anaethelion (changes requested):
Overall seems like a good change.
A few questions related to error handling of flushes.

esutil/bulk_indexer.go
rockdaboot: I simply kept the logic as is, which is calling `bi.config.OnError()` if `flush()` returns an error. There is only one case where the flush error return is needed (see L407).
Indeed we could return `bool` instead of `error`. Though `error` might be useful in the future 🤷🏻‍♂️ .

Let me know what you prefer.

Anaethelion: Let's make that a boolean for the sake of consistency, we can always add back the error in the future if we need it.

Anaethelion (approved):

Anaethelion:
macOS test has been updated in between and is not related to this PR, nor is the clients-ci.
LGTM!

PR #657, opened 2023-04-11
Reties With not empty URL Path

f-dg:
transport builds an incorrect request path if the address is configured with a path, for example:

```
	eCnf := elasticsearch.Config{
		Addresses: []string{
			"https://server.com/somepath"
		},
	}
``` 
it happens when retries is enabled and the API returns an error, transports adds to the request "/somepath" on each retry instead of building a new path.

### DIFF
--- a/estransport/estransport.go
+++ b/estransport/estransport.go
@@ -327,6 +327,8 @@ func (c *Client) Perform(req *http.Request) (*http.Response, error) {
 			}
 		}
 	}
+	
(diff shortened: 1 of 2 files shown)


f-dg:
CA is signed, but I don't see any triggers to rerun the check

Anaethelion:
Hi @f-dg 

Thank you for noticing this, LGTM!
I'll push a minor change and adapt a test to reflect the fix.

PR #665, opened 2023-05-03
Read document and mapper from file

Adeesh2411:
## Support of document.json and mapping.json for indexing
Added Support to index the documents from a json file and have mapping support from the file itself, instead of hardcoding the go file. This feature is present in the later versions, but since most of us still using 6.x, thought this could be of use.


Anaethelion:
Thank you @Adeesh2411 for this, unfortunately this client only supports the latest version of Elasticsearch and the upmost version of the previous, which is `7.17`.

PR #694, opened 2023-07-09
chore: use String() instead of fmt.Sprintf

testwill:


Anaethelion:
Hi there,

The code you modified is generated and cannot be merged as-is.
While this would be an improvement the code generator doesn't yet handle that kind of detection.

If you encountered an issue with the serialization of those object, please open an issue.

PR #701, opened 2023-07-19
Fix memory leak bulk indexer

mhmtszr:
Bulk indexer makes a lot of heap allocation, it affect our applications' performance. I tried to reduce allocations by using "sync.pool"

### DIFF
--- a/esutil/bulk_indexer.go
+++ b/esutil/bulk_indexer.go
@@ -251,7 +251,8 @@ type bulkIndexer struct {
 	workers []*worker
 	stats   *bulkIndexerStats
 
-	config BulkIndexerConfig
+	config  BulkIndexerConfig
+	bufPool *sync.Pool
 }
 
 type bulkIndexerStats struct {
@@ -290,6 +291,11 @@ func NewBulkIndexer(cfg BulkIndexerConfig) (BulkIndexer, error) {
 	bi := bulkIndexer{
 		config: cfg,
 		stats:  &bulkIndexerStats{},
+		bufPool: &sync.Pool{
+			New: func() interface{} {
+				return bytes.NewBuffer(make([]byte, 0, cfg.FlushBytes))
+			},
+		},
 	}
 
 	bi.init()
@@ -337,6 +343,11 @@ func (bi *bulkIndexer) Close(ctx context.Context) error {
 		bi.wg.Wait()
 	}
 
+	for _, w := range bi.workers {
+		w.buf.Reset()
+		bi.bufPool.Put(w.buf)
+	}
+
 	return nil
 }
 
@@ -363,7 +374,7 @@ func (bi *bulkIndexer) init() {
 			id:     i,
 			ch:     bi.queue,
 			bi:     bi,
-			buf:    bytes.NewBuffer(make([]byte, 0, bi.config.FlushBytes)),
+			buf:    bi.bufPool.Get().(*bytes.Buffer),
 			ticker: time.NewTicker(bi.config.FlushInterval),
 		}
 		w.run()


T-J-L:
What's the reason for opening/closing indexers? It can last the lifetime of the application, when doing so with a large buffer there are zero allocations here.

mhmtszr:
@T-J-L hello, is there any way to use same indexers for different operations? we needed to close after adding to batch per operation.

T-J-L:
> @T-J-L hello, is there any way to use same indexers for different operations? we needed to close after adding to batch per operation.

I create a single indexer at start up, with a low flush time (100ms). Then for each application request create a couple of channels for success/errors, perform `BulkIndexer.Add` then write back to the channels in the `BulkIndexerItem.OnSuccess` and `BulkIndexerItem.OnFailure` callbacks. So effectivly each request is syncronous, with all requests to ES are batched.

You can set the `Index` and `Action` per item, so this works for all types of operation.

mhmtszr:
> > @T-J-L hello, is there any way to use same indexers for different operations? we needed to close after adding to batch per operation.
> 
> I create a single indexer at start up, with a low flush time (100ms). Then for each application request create a couple of channels for success/errors, perform `BulkIndexer.Add` then write back to the channels in the `BulkIndexerItem.OnSuccess` and `BulkIndexerItem.OnFailure` callbacks. So effectivly each request is syncronous, with all requests to ES are batched.
> 
> You can set the `Index` and `Action` per item, so this works for all types of operation.

Great solution, but how can you be sure your documents will be written to Elasticsearch? We need to be sure that our documents will be written to Elasticsearch thus we are closing the bulk indexer.

JAndritsch:
> > @T-J-L hello, is there any way to use same indexers for different operations? we needed to close after adding to batch per operation.
> 
> I create a single indexer at start up, with a low flush time (100ms). Then for each application request create a couple of channels for success/errors, perform `BulkIndexer.Add` then write back to the channels in the `BulkIndexerItem.OnSuccess` and `BulkIndexerItem.OnFailure` callbacks. So effectivly each request is syncronous, with all requests to ES are batched.
> 
> You can set the `Index` and `Action` per item, so this works for all types of operation.

It would seem odd to me that the intended use of a `BulkIndexer` would be a singleton instance living for the lifetime of an application. Especially since the examples in this repo show calling `Close()` on the indexer after your operations are done.

As @mhmtszr suggested, `Close()` will force a flush and wait for the documents to index. That should also theoretically free up any resources used by the `BulkIndexer`, but it seems that's not the case.

I will try to test these changes and see if it resolves my issue. I'm not a maintainer of the repo, but hopefully that will help with the acceptance of this PR.

JAndritsch:
I can confirm that this change resolves the memory issues I've been seeing: https://github.com/elastic/go-elasticsearch/issues/956.

JAndritsch (approved):
I tested these changes in my own custom fork of the repo, since the original clone here is a bit old. 

Prior to these changes, memory usage would spike when calling `Close()` on multiple instances of `BulkIndexer`. My application would spike from ~8mb to ~300mb of usage after just a few invocations.

I saw memory usage remain much more expected after applying these changes to my fork. Memory usage stayed within ~1-2mb of the original bootup usage.

Resolves https://github.com/elastic/go-elasticsearch/issues/956

Edit: Although I'm seeing greatly improved memory usage after this change, I'm starting to wonder if this fix masks an underlying problem with how resource cleanup is handled in `BulkIndexer`.  The `flushBuffer` method is also allocating new buffers in a similar way that the `NewBulkIndexer` was prior to this PR change.

I think there's probably more to investigate here.

T-J-L:
> > > @T-J-L hello, is there any way to use same indexers for different operations? we needed to close after adding to batch per operation.
> 
> > 
> 
> > I create a single indexer at start up, with a low flush time (100ms). Then for each application request create a couple of channels for success/errors, perform `BulkIndexer.Add` then write back to the channels in the `BulkIndexerItem.OnSuccess` and `BulkIndexerItem.OnFailure` callbacks. So effectivly each request is syncronous, with all requests to ES are batched.
> 
> > 
> 
> > You can set the `Index` and `Action` per item, so this works for all types of operation.
> 
> 
> 
> It would seem odd to me that the intended use of a `BulkIndexer` would be a singleton instance living for the lifetime of an application. Especially since the examples in this repo show calling `Close()` on the indexer after your operations have been added.
> 
> 
> 
> As @mhmtszr suggested, `Close()` will force a flush and wait for the documents to index. That should also theoretically free up any resources used by the `BulkIndexer`, but it seems that's not the case.
> 
> 
> 
> I will try to test these changes and see if it resolves my issue. I'm not a maintainer of the repo, but hopefully that will help with the acceptance of this PR.

IMO a long running process is the only reason for using the bulk indexer. If you are closing the indexer immediately to force a flush, it seems like using a normal bulk request would be a better choice.

MattDevy:
Thanks for raising this, @mhmtszr, and for the investigation and confirmation from @JAndritsch.

We looked into the buffer pool approach here but decided against merging it. Adding a `sync.Pool` for the internal buffers can introduce its own issues: pooled buffers tend to grow to the high-water mark and stay there, the pool's interaction with GC timing is unpredictable, and it changes the memory ownership semantics of the workers in ways that are hard to reason about under concurrent use.

Instead, we've addressed the underlying problem differently. The core issue is that users were creating and closing `BulkIndexer` instances repeatedly (as shown in #956), when the indexer is designed to be long-lived. The repeated create/close cycle allocates large internal buffers each time, which is where the memory pressure comes from.

As of the following releases, we've added a public `Flush(ctx context.Context) error` method to `BulkIndexer`:

- [v8.19.4](https://github.com/elastic/go-elasticsearch/releases/tag/v8.19.4)
- [v9.2.4](https://github.com/elastic/go-elasticsearch/releases/tag/v9.2.4)
- [v9.3.2](https://github.com/elastic/go-elasticsearch/releases/tag/v9.3.2)

`Flush` drains all queued items and sends them to Elasticsearch, then returns. The indexer remains fully usable afterward, so you can keep a single instance for the lifetime of your application and call `Flush` whenever you need the guarantee that queued documents have been sent. This avoids the repeated allocation overhead entirely.

For the pattern @mhmtszr described (needing to ensure documents are written per operation), the recommended approaches are:

1. **Long-lived indexer + `Flush`**: Create the indexer once, call `Add` as needed, and call `Flush` when you need confirmation that items have been dispatched. Per-item results still come through `OnSuccess`/`OnFailure` callbacks.
2. **Direct bulk request**: If you're batching a known set of documents and want a synchronous response, use the `Bulk` API directly rather than the `BulkIndexer`, which is designed for async, high-throughput use cases.

For full usage guidance and examples, see the [Bulk indexing documentation](https://www.elastic.co/docs/reference/elasticsearch/clients/go/using-the-api/bulk-indexing).

Closing this in favor of the `Flush` approach. Thanks again for the contribution and the discussion.

PR #715, opened 2023-08-29
Unmarshal NaN and Infinite values correctly

fisherevans:
The current implementation of unmarshaling Float64s does not handle Infinity values. This can be tested by creating a Sum aggregate of two document attributes that are both "Max Float" values or "Negative Max Float" values. The current behavior is a bubbled up error from `strconv.ParseFloat(string(data), 64)` which cannot handle a value like `"+Infinity"`.


fisherevans:
Signed

Anaethelion (approved):
Thank you!

I'll merge this for proper credit and will backport to the generator for future propagation.

PR #716, opened 2023-09-01
Resolve a deadlock for v6.8.10, same as elastic#603

hidu:
same as #603


Anaethelion:
Thank you for this!
Unfortunately the 6.x branch is past [EOL](https://www.elastic.co/support/eol) and is not released anymore.

PR #727, opened 2023-09-28
Fix positive infinity unmarshalling

xlucas:
Unfortunately #715 contained a subtle error because the representation of positive infinity in JSON as returned by Elasticsearch is `"Infinity"` and not `"+Infinity"`.


Anaethelion:
Hi! Thank you for the report, I've fixed this in #731

PR #729, opened 2023-10-09
utils/map: preallocate memory

florianl:
Preallocate memory for result of MapKeys and MapValues to avoid expensive reallocations and further GC actions.


Anaethelion (approved):
LGTM!

PR #824, opened 2024-03-06
feat: use testcontainers-go for the integration tests

mdelapenya:
Hi :wave: from Testcontainers Go community! I'm sending this PR as way to improve the developer experience running the tests in this project, using testcontainers-go.

Of course, it's up to the maintainers whether to include it or not, and I'm happy with any of them, although I'd like to see it included for the reasons explained in the `Why is it important?` of this PR.

## What does this PR do?
This PR adds [testcontainers-go](https://github.com/testcontainers/testcontainers-go) as a means to start Elasticsearch containers while running the integration tests.

I'm only adding it to one test function to demonstrate the benefits of declaring the dependencies in the code, compared to use a big set of shell scripts to start the test-time dependencies.

The `containertest` internal package could be used across different test files containing integration tests, and won't be included in any final binary if not imported from any non-test file. 

## Why is it important?
It will help contributors to run the integration tests with more ease: just cloning the repo and running `make test-integ`. The same would apply to the CI.

The code is also self-contained, which means that there is no need to start containers in the CI (Buildkite) in a way that developers need to learn. Here, the dependencies are declared programatically next to the tests, so it's easier to maintain.

Besides that, the plumbing scripts will get reduced, as everything will be done in code, which is easier to test.

## Other considerations
I did not configured the Elasticsearch instance as described in the CI shell scripts, and it was on purpose, to showcase how it's possible to start the ES instance with a few lines of code. All the configurations in there can be added to the module thanks to the module configuration API. Please see https://golang.testcontainers.org/modules/elasticsearch/ for more information about the module.

I'm not removing any shell script code (yet), as I'm not that familiar with the codebase, but this PR would be an initial step of a few of them. For that, I'd thank any feedback from your review.

Cheers!


internal/containertest/testcontainers.go
```diff
@@ -0,0 +1,56 @@
...
+
+	"github.com/elastic/go-elasticsearch/v8/internal/version"
+
+	"github.com/testcontainers/testcontainers-go"
+	tces "github.com/testcontainers/testcontainers-go/modules/elasticsearch"
+)
+
+// SetupElasticsearch starts an Elasticsearch container and sets the environment variables
+// ELASTICSEARCH_URL, ELASTICSEARCH_USERNAME, and ELASTICSEARCH_PASSWORD in the test context.
+func SetupElasticsearch(t *testing.T) (*tces.ElasticsearchContainer, error) {
```
Anaethelion, line 36: I would wrap this in a structure which could both allow to stop the container once the test has ended and would return the base config.

This would avoid recasting the config for each requests and would allow to reuse that in `esutil` and `_examples`.

wdyt ?

Anaethelion (changes requested):
Hi @mdelapenya glad to see you here!

Thank you for the great write-up about testcontainers. I've been meaning to use it for some time, this is the perfect occasion to do that.
I agree that integration tests are a good fit for a first step!

However the buildkite shell script is here to stay as not all clients have access to a testcontainers implementation and this is the cornerstone of our mutual compatibility within the team.

That said, integration tests in the client, bulkindexer and several examples could still profit from this.

I've left a question on which we can iterate other than that it looks good!

mdelapenya:
> That said, integration tests in the client, bulkindexer and several examples could still profit from this.

Yeah, sure. I do not have the knowledge of the codebase to know where it testcontainers-go fits the best, so whatever you see, fine for me 🙋

internal/containertest/testcontainers.go
mdelapenya: Sounds good to me. I tried to address it in the latest commits

mdelapenya:
@Anaethelion I've resolved the conflicts here. Feel free to ping me for anything you need with this PR.

Cheers!

Anaethelion:
@mdelapenya Sorry this has been put on the back-burner.

Picking this up I've ultimately decided that I don't want users to tract the test dependencies by simply using the lib.
I'm going to take over that PR and move things around a bit to push the integration testing in a private package in `internal`.

Won't change much on what has been done so far, just moving things around!

Anaethelion (approved):
Thank you! :rocket: 

For the snyk related error: this will be augmented by #855

PR #827, opened 2024-03-14
Bump to go1.21

xeviknal:
Hello! It's the first time I am contributing here :)

Let me know what I can do to complete this PR.

Thanks!


Anaethelion (changes requested):
Thank you for this, I was going to create the same one.
For this to be complete the examples should be bumped as well.

Anaethelion (approved):
That was slightly more complicated than it looked at first sight so I took over.
I hope you don't mind, you'll still get credited of course. :tada:

PR #840, opened 2024-04-01
add `PoolCompressor` option and propagate it to `elastictransport.Config`

pakio:
This is a follow up PR for https://github.com/elastic/elastic-transport-go/pull/19

---- 

Added `PoolCompressor` option into `elasticsearch.Config` and pass it into `elastictransport.Config`.


Anaethelion (approved):
LGTM!

PR #885, opened 2024-07-04
Fix go directive version in go.mod

mauri870:
According to https://go.dev/issues/62278 a concrete go version has to include the patch number or an additional toolchain directive. The project declares the version as 1.21 which is invalid and will result in an error:

	GOTOOLCHAIN="go1.20+auto" go1.21.5 mod tidy
	go: downloading go1.21 (linux/amd64)
	go: download go1.21 for linux/amd64: toolchain not available

Changed to 1.21.0 as this is the first actual release of go 1.21.


Anaethelion:
Thank you! I'll update the relevant go.mod files to make sure the CI is up to date.

Anaethelion (approved):
CI is fixed, Snyk will be dealt with separately.
Thanks again!

Issue #796, opened 2024-02-05
BulkIndexer WaitGroup delta is added after workers are launched


Jakub-Kapusta:
https://github.com/elastic/go-elasticsearch/blob/a58e89f1ecc47a94db81a93c13935e8da3b8438b/esutil/bulk_indexer.go#L380-L383

The `wg` Delta should generally be added before relevant go-routines are launched.
Reference: https://pkg.go.dev/sync#WaitGroup.Add

This might not be a big issue in this case, but it strays from best practices.
Also, it creates a small window where a worker could call `Done()` before the call to `Add()`, which is not ideal.

PR #891, opened 2024-07-24
Bulk indexer: Follow convention and increment waitgroup in loop before creating goroutines

HaraldNordgren:
The wg Delta should generally be added before relevant go-routines are launched. Reference: https://pkg.go.dev/sync#WaitGroup.Add

This also aligns the code with our other uses of waitgroups.

Closes https://github.com/elastic/go-elasticsearch/issues/796.


HaraldNordgren:
Ping @Jakub-Kapusta

HaraldNordgren:
Ping @Anaethelion

Anaethelion (approved):

PR #908, opened 2024-09-26
Remove duplicate and redundant code

cuishuang:


HaraldNordgren (changes requested):
This appears to be generated code. I think this is not the correct place to remove it from.

MattDevy:
Hi @cuishuang 

I agree with @HaraldNordgren, this is generated code and shouldn't be edited directly. If you could provide more context for why these changes are needed I can look into the code generation to improve it.

PR #914, opened 2024-10-29
Expose BulkIndexer total flushed bytes metric

aureleoules:
This pull request exposes the amount of bytes flushed in total by the Bulk indexer.

I tested this change locally and it worked as expected.

### DIFF
--- a/esutil/bulk_indexer.go
+++ b/esutil/bulk_indexer.go
@@ -86,14 +86,15 @@ type BulkIndexerConfig struct {
 
 // BulkIndexerStats represents the indexer statistics.
 type BulkIndexerStats struct {
-	NumAdded    uint64
-	NumFlushed  uint64
-	NumFailed   uint64
-	NumIndexed  uint64
-	NumCreated  uint64
-	NumUpdated  uint64
-	NumDeleted  uint64
-	NumRequests uint64
+	NumAdded     uint64
+	NumFlushed   uint64
+	NumFailed    uint64
+	NumIndexed   uint64
+	NumCreated   uint64
+	NumUpdated   uint64
+	NumDeleted   uint64
+	NumRequests  uint64
+	FlushedBytes uint64
 }
 
 // BulkIndexerItem represents an indexer item.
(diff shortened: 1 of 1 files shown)


esutil/bulk_indexer.go
```diff
@@ -508,7 +511,9 @@ func (w *worker) flushBuffer(ctx context.Context) error {
 		defer func() { w.bi.config.OnFlushEnd(ctx) }()
 	}
 
-	if w.buf.Len() < 1 {
+	bufLen := w.buf.Len()
+
+	if bufLen < 1 {
 		if w.bi.config.DebugLogger != nil {
```
jbartosik, line 517:
```suggestion
	if bufLen < 1 && w.bi.config.DebugLogger != nil {
```

esutil/bulk_indexer.go
```diff
@@ -631,6 +636,8 @@ func (w *worker) flushBuffer(ctx context.Context) error {
 		}
 	}
 
+	atomic.AddUint64(&w.bi.stats.flushedBytes, uint64(bufLen))
```
jbartosik, line 639: Nit: I'd move it above the if, for me that's easier to read

esutil/bulk_indexer.go
aureleoules: This won't translate to the same code as before because of the `return nil` circuit break below

esutil/bulk_indexer.go
aureleoules: Do you mean above the for loop?

esutil/bulk_indexer.go
jbartosik: Right

esutil/bulk_indexer.go
jbartosik: I meant something like line 522, back where `bufLen` is set. But  I missed that there are a bunch of returns in the meantime so we might not get to this line. Let's keep as is

jbartosik (commented):
Looks good to me

mfpierre (approved):

mfpierre:
👋 Hey @Anaethelion,
could you have a look at this one? Exposing this stat in the client would be really helpful for us, thanks 🙏

Anaethelion (approved):

PR #948, opened 2025-01-31
esutil: add failure_store key to bulk response item

1pkg:
# Description

This PR updates `BulkIndexerResponseItem` object to include `failure_store` key. This is a dependency for changes made in https://github.com/elastic/go-docappender/pull/228.


axw (approved):
LGTM. For others' reference, see https://github.com/elastic/elasticsearch/blob/1225b0720aba89247c14727bf2e8e29a1af96cd0/server/src/main/java/org/elasticsearch/action/bulk/IndexDocFailureStoreStatus.java

PR #1032, opened 2025-10-22
fix: replace any with json.RawMessage in MarshalJSON of types.Query t…

### LINKED ISSUE #1021: types.Query truncates int64 during marshaling
`types.Query` truncates `int64` in its custom `MarshalJSON` implementation.

Probably happens because [here](https://github.com/alexartwww/go-elasticsearch/blob/8970b63b259e/typedapi/types/adjacencymatrixbucket.go#L516) you marshal Query into JSON and then turn it into map[string]any. `json` package probably convert int64 into float and this causes loss of precision. 

As a result this query fails for me because `tags` value changes.

```json
{
    "from": 0,
    "query": {
        "bool": {
            "must": [
                {
                    "terms": {
                        "tags": [
                            4475034517144338534
                        ]
                    }
                }
            ]
        }
    },
    "size": 20
}
```
4475034517144338**534** gets truncated to 4475034517144338**400**

<details>
<summary>demo of the issue</summary>
The following is analogous to what happens in `MarshalJSON` of `types.Query`

```golang
package main

import "encoding/json"

type Issue1021 struct {
	Value int64
}

func main() {
	v := Issue1021{
		Value: 4475034517144338534,
	}
	jsonV, _ := json.Marshal(v)
	println(string(jsonV))
	var tmp map[string]any
	json.Unmarshal(jsonV, &tmp)
	jsonV2, _ := json.Marshal(tmp)
	println(string(jsonV2))
}
```
Output:
```
{"Value":4475034517144338534}
{"Value":4475034517144338400}
```
</details>

MaratBR:
…o prevent unexpected chat conversions and precision loss

This fixes #1021

### DIFF
--- a/typedapi/types/query.go
+++ b/typedapi/types/query.go
@@ -617,7 +617,7 @@ func (s *Query) UnmarshalJSON(data []byte) error {
 func (s Query) MarshalJSON() ([]byte, error) {
 	type opt Query
 	// We transform the struct to a map without the embedded additional properties map
-	tmp := make(map[string]any, 0)
+	tmp := make(map[string]json.RawMessage, 0)
 
 	data, err := json.Marshal(opt(s))
 	if err != nil {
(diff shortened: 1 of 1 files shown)


MattDevy:
Hi @MaratBR thank you for this contribution!

I have made adjustments in the code generator to fix this issue. 

Please can you let me know what version of the library you are using as I intend to backport this fix to:

- 9.2
- 9.1
- 8.19

Let me know if you're using a different version of the go-elasticsearch library

PR #1033, opened 2025-10-23
Add custom transport

khushijain21:
This PR adds a new configuration parameter `CustomTransport` that replaces the default implementation of `elastictransport.Interface` https://github.com/khushijain21/go-elasticsearch/blob/customtrans/elasticsearch.go#L140

This is useful, in particular to support keberos authentication which must be performed at application (or client) level.

### DIFF
--- a/elasticsearch.go
+++ b/elasticsearch.go
@@ -21,7 +21,6 @@ import (
 	"encoding/base64"
 	"errors"
 	"fmt"
-	"go.opentelemetry.io/otel/trace"
 	"net/http"
 	"net/url"
 	"os"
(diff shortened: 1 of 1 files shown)


elasticsearch.go
```diff
@@ -258,7 +278,7 @@ func NewTypedClient(cfg Config) (*TypedClient, error) {
 	return client, nil
 }
 
-func newTransport(cfg Config) (*elastictransport.Client, error) {
+func NewTransport(cfg Config) (*elastictransport.Client, error) {
```
khushijain21: I make this a public method to allow wrapping Kerberos logic around this in a separate integration

khushijain21:
Hello @MattDevy, this sounds good. I'll close this PR

Issue #391, opened 2021-12-26
NewBulkIndexer never returns an error


liamphmurphy:
Hi all, 

I've been messing around with this library and came across that [NewBulkIngestor](https://github.com/elastic/go-elasticsearch/blob/6cdebcf1b94eb2bf022ce0689762049cb07103e0/esutil/bulk_indexer.go#L188) does not seem to return an error in any case. The only place where an error could be obtained, when creating the Elastic client if it is nil, the error message is completely ignored. 

I had a few thoughts;

1. If that function is never going to return an error, the func signature should change so the user isn't expected to handle an error that will never occur
2. I personally don't understand the reason why creating a default Elastic client is desirable if it is nil, I would expect that to be an error along the lines of "elastic client is nil" so that the user is told to handle the making of a new client. A bulk indexer init func, in my mind, should not make that decision.

Happy to make whatever change is decided. In my opinion, we should forego the making of that default client (unless there is a use case I'm missing) and return an error in the case that `cfg.Client == nil`. 

Thanks all for your time.

MattDevy:
This is resolved. As of v9.4, `NewBulkIndexer` no longer silently creates a default client. `BulkIndexerConfig.Client` is required, and a nil value returns an error:

```
BulkIndexerConfig.Client is required
```

The implicit fallback path that swallowed the error has been removed entirely.

PR #1302, opened 2026-03-05
fix(esutil): propagate error from NewDefaultClient in NewBulkIndexer

amitkanfer:
## Summary

- `NewBulkIndexer` silently fell through to a nil-pointer dereference when `NewDefaultClient()` failed (e.g. invalid `ELASTICSEARCH_URL`). The error is now propagated to the caller.

## Changes

- `esutil/bulk_indexer.go`: return the error from `elasticsearch.NewDefaultClient()` instead of silently proceeding with a nil client.
- `esutil/bulk_indexer_internal_test.go`: added `TestNewBulkIndexerNilClientError` that sets `ELASTICSEARCH_URL` to an invalid value and asserts a non-nil error is returned.

Fixes https://github.com/elastic/go-elasticsearch/issues/391

🤖 Generated with [Claude Code](https://claude.com/claude-code)


MattDevy:
Thanks for the contribution. As of v9.4, `BulkIndexerConfig.Client` is now required and `NewBulkIndexer` returns an error if it's nil, so the implicit `NewDefaultClient()` fallback no longer exists. This makes the change here superseded.

Closing, but this also resolves the underlying issue #391.

PR #1335, opened 2026-03-30
fix!: BulkIndexer requires an explicit Client

### LINKED ISSUE #462: BulkIndexer throw error when es.Client not provided ?
Currently, NewBulkInderxer function creates a elasticsearch.Client with default ES URL=localhost:9200 in case if it is not provided by the consumer. In most of the PROD ready applications ES Url address would be different from the default address and it makes sense to give a feedback in terms of error rather than creating it to default. This will help in getting an early feedback to the consumer to course correct it.
https://github.com/elastic/go-elasticsearch/blob/main/esutil/bulk_indexer.go#L190

thani-ath-nain:
Addresses https://github.com/elastic/go-elasticsearch/issues/462 so that instead of silently making the default client if client is nil, an error is returned to let the user know

BEGIN_COMMIT_OVERRIDE
chore: BulkIndexer requires an explicit Client
END_COMMIT_OVERRIDE


MattDevy (changes requested):
Thank you for this contribution! (I fixed the CI issue, apologies for that)

I've added a few suggestions, given that the error explains itself well, I would suggest no comment is needed.

Additionally, adding the struct name in the error will help users identify the cause if the error is deeply nested in their codebase.

Since this will be a breaking change for any user that currently does not pass a client, I will not backport this PR to the older version(s) of the client. But it will be included in all future versions :)

MattDevy (approved):
LGTM! Thank you!

PR #1386, opened 2026-04-16
feat(esutil): add FlushJitter option to BulkIndexer

### LINKED ISSUE #646: BulkIndexer: Workers flushing at the same time
## The Problem
The workers tend to flush at (roughly) the same time.

The main reason for this is that worker buffers are filled evenly, because all workers fetch their items from the same Go channel in parallel.

The buffer expiration made it worse, because so far (up to go-elasticsearch 8.7), there is one ticker that flushes all workers at the same time. #624 fixed this for 8.8+.

The flushing at the same time has several bad effects:
- peak in memory usage: the bulk indexer items are kept in memory until flush, the buffer memory (HTTP body) is allocated/filled at the same time
- peak in CPU consumption: the HTTP bodies are generated and compressed at the same time
- peak in network usage: all requests to ES go out in parallel
- same peak effects on the Elastic Search side, leading to 429 responses more often than needed (and expected), consequentially leading to more retries, which amplifies the above peak behavior

## Possible solution: Fill buffers sequentially, flush in background
(Out of scope is changing the API)

Let's say we have number of workers set to N with `FlushBytes` B and FlushInterval I.
Let the Add() function collect items into an array A0 until B or I is reached, then flush in the background.
Further calls to Add() are going into a new array A1 until B or I is reached, then flush in the background.
...
Allow maximal N background flushes. If reached, throttle ingestion in the Add() function (as we do now).

Pros:
- spread the workload over time to reduce peak behavior and pressure on ES
Cons:
- ?

MattDevy:
## Summary

Closes #646.

`esutil.BulkIndexer` runs `NumWorkers` goroutines, each with its own `time.Ticker`. Because:

- all tickers are created at the same instant in `bulkIndexer.init()`,
- round-robin dispatch fills worker buffers evenly, and
- `ticker.Reset()` after every flush re-synchronizes the cycle,

workers tend to flush in lockstep, producing bursty load on Elasticsearch instead of evenly-spaced bulk requests. This is the behavior #646 originally surfaced. Per-worker tickers (added in #624) removed the shared-ticker contention but didn't solve the synchronization problem.

The original proposal (single ingestion buffer + N background flushers) would reintroduce contention on `Add()` and conflict with the channel-based explicit `Flush()` design landed in #1346. This PR takes the jitter approach suggested in [this comment](https://github.com/elastic/go-elasticsearch/issues/646#issuecomment-...):

- Adds `BulkIndexerConfig.FlushJitter time.Duration`.
- When `FlushJitter > 0`, each worker uses `FlushInterval + rand.N(FlushJitter)` for both `time.NewTicker` at init and `ticker.Reset` after every flush.
- Defaults to `0`, preserving current behavior byte-for-byte.

The randomness source is `math/rand/v2` (goroutine-safe top-level funcs, no seeding).

## Implementation

- `esutil/bulk_indexer.go`: new `FlushJitter` field on `BulkIndexerConfig`, new `(*bulkIndexer).nextFlushInterval()` helper, two call-site swaps (init + post-flush reset).
- `esutil/bulk_indexer_internal_test.go`: unit tests for the helper (zero / negative / bounds across 1000 samples) and an integration-style test that verifies an auto-flush still fires with jitter configured. Also a guard test that the default remains `0`.

Docs are split out into #1438 so this branch backports cleanly to version branches (which do not carry `docs/reference/**`).

No changes to generated code, examples, or benchmarks. No API breakage.

## Test plan

- [x] `make lint`
- [x] `make test-unit` — 317 tests pass
- [x] `go test -race ./esutil/` — green
- [x] `go test -race -run 'TestBulkIndexerNextFlushInterval|TestBulkIndexerFlushJitter' -v ./esutil/` — 5/5 green
- [ ] Manual/benchmark sanity check that per-second flush counts smooth out with jitter set (optional; not a gate).

Draft for review of the approach before we commit to the config field name / surface.


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/clients-team/pr-checks/bb9b4eb1-61f5-4ce5-91fd-cc165cbc9102?refs=6cad01a8fd3a27ae6ade517d33d2ca3bf1aec164&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/clients-team/pr-checks/bb9b4eb1-61f5-4ce5-91fd-cc165cbc9102/license?refs=6cad01a8fd3a27ae6ade517d33d2ca3bf1aec164&source=prChecksComment) |

Anaethelion (approved):

Issue #52, opened 2019-05-13
Cannot use forward slash in document ID


bpmoneill:
# Issue
When adding a document to Elasticsearch (v7.0.1) using go-elasticsearch on master branch, it is not possible to use the '/' as a separator within the document ID field e.g. title/test.txt.

Sample code:
```
func example() {
	client, _ := es.NewDefaultClient()
	docID := "title/test.txt"
	req := esapi.IndexRequest{
		Index:      "test",
		DocumentID: docID,
		Body:       strings.NewReader(`{"title" : "test.txt"}"`),
		Refresh:    "true",
	}

	res, _ := req.Do(context.Background(), client)
	defer res.Body.Close()

	if res.IsError() {
		log.Printf("[%s] Error indexing document ID=%s", res.String(), docID)
	}
}
```
Using curl I can insert the document by manually encoding the '/' as below:
```
curl -X PUT "localhost:9200/test/_doc/title%2Ftest" -H 'Content-Type: application/json' -d' 
{ 
    "title" : "test.txt" 
} 
' 
```

# Expected
200 OK with document added to Elasticsearch.  

# Actual
Error produced as below:
```
[[400 Bad Request] {"error":"no handler found for uri [/test/_doc/title/test.txt?refresh=true] and method [PUT]"}] Error indexing document ID=title/test.txt
```

# Root Cause?
It would seem from looking at the code that the URL is not constructed correctly.  Within `esapi.request.go` it constructs the request as below:
```
func newRequest(method, path string, body io.Reader) (*http.Request, error) {
	r := http.Request{
		Method:     method,
		URL:        &url.URL{Path: path},
		Proto:      "HTTP/1.1",
		ProtoMajor: 1,
		ProtoMinor: 1,
		Header:     make(http.Header),
	}
  ...
```
The path, specifically the document ID part, does not get encoded and the URL structure is not initialised correctly, meaning the necessary RawPath field is not set.  In order to resolve this issue, the document ID would need to be encoded prior to use and the URL should be initialised using the `url.Parse(path)` correctly setting Path and RawPath fields.

karmi:
Hi, thanks for the report!

I wouldn't like to use `url.Parse()` because I'm worried about performance implications in such a hot path, and, moreover, the `newRequest()` function is probably too late to do the escaping: that would escape the _whole_ URL, including eg. the slashes which shouldn't be escaped.

[...]

PR #1425, opened 2026-04-16
fix(gen): percent-escape user-supplied path parts in esapi generator

### LINKED ISSUE #183: index name may be double escaped
In Elasticsearch, escaped characters may be contains in index name.
ref: https://www.elastic.co/guide/en/elasticsearch/reference/current/date-math-index-names.html

```
# PUT /<my-index-{now/d}>
PUT /%3Cmy-index-%7Bnow%2Fd%7D%3E
```

Notice that we have to escape the `/` here.

In Golang, paths are usually escaped **automatically**. However, `/` is **not** an escaped character.
If we specify an escaped index name, it will be double escaped.

So, the library is currently not possible to use this index.

-----

Solution:
https://github.com/soranoba/go-elasticsearch/compare/master..fix/master/double-escape-path

The code will solve the problem, but what was previously escaped will no longer be escaped.
It may be a breaking change.

Is there a better solution?

MattDevy:
User-supplied path parts (index names, document IDs, node IDs) are
currently written raw into the path strings.Builder passed to
http.NewRequest. Go's net/url auto-escapes most reserved characters but
deliberately leaves '/' alone because it treats the string as a full
path, so requests whose index name or ID contains '/' break — most
notably date-math names like '<my-index-{now/d}>', which must hit the
server as /%3Cmy-index-%7Bnow%2Fd%7D%3E. Caller-side pre-escaping also
fails because a literal '%' in URL.Path is re-escaped to %25.

Mirror the approach taken in the typedapi generator
(elastic/elastic-client-generator-go#141): maintain a dual path /
rawPath strings.Builder in each generated Do method. User segments are
written raw to path and through a new escapePathPart helper to rawPath;
static segments and separators are written to both identically. After
newRequest succeeds, assign req.URL.Path and req.URL.RawPath from the
two builders so the transport serializes the escaped form.

escapePathPart uses url.PathEscape and preserves commas so
comma-joined list path params keep their element separators while
each element is independently escaped.

This PR contains only the generator change + the hand-written helper.
The mechanical esapi/ regeneration will land in a follow-up.

Closes #183
Refs #52
Refs elastic/elastic-client-generator-go#141

### DIFF
--- a/esapi/esapi.escape.go
+++ b/esapi/esapi.escape.go
@@ -1,0 +1,38 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Elasticsearch B.V. licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package esapi
+
+import (
+	"net/url"
+	"strings"
+)
+
+// escapePathPart percent-escapes a single URL path segment while preserving
+// commas. Commas act as separators inside comma-joined list path params
+// (e.g. "index1,index2"), so each element is escaped independently and the
+// commas are re-joined verbatim.
+func escapePathPart(s string) string {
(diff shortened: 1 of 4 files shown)


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/clients-team/pr-checks/f5fd7b7b-f082-445a-ace4-04e21269bb0c?refs=23a6484ef196dfb28c363569d1a8c84b866b55bd&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/clients-team/pr-checks/f5fd7b7b-f082-445a-ace4-04e21269bb0c/license?refs=23a6484ef196dfb28c363569d1a8c84b866b55bd&source=prChecksComment) |

PR #1451, opened 2026-04-21
docs(examples): migrate _examples/** to TypedClient + esdsl

MattDevy:
## Summary

Continues the push to favour `TypedClient` + `esdsl` over the functional `esapi` client. Migrates the in-scope example apps under `_examples/` to the typed API, converts the top-level customization demo to the Interceptor pattern, and retires two legacy third-party instrumentation demos (Elastic APM v1, OpenCensus) that predate Interceptors and the OTel-via-interceptor demo we already ship.

## What changed

- **Top-level getting-started** — `_examples/main.go`, `configuration.go`, `customization.go` switch to `elasticsearch.NewTyped()` + `Info().Do(ctx)`, `Index(idx).Document(...).Refresh(refresh.True)`, and `Search().Query(esdsl.NewMatchQuery(...))`. `customization.go` additionally drops its custom `CountingTransport` (http.RoundTripper) in favour of an Interceptor closure that adds headers, logs, and counts requests.
- **xkcdsearch app** — `store.go`, `store_test.go`, and the three `cmd/xkcd/commands/*` files. `StoreConfig.Client` is now `*elasticsearch.TypedClient`; `Create`, `Exists`, `CreateIndex`, and `Search` use the typed builders. Search uses `esdsl.NewMultiMatchQuery`, `esdsl.NewHighlight`, and `esdsl.NewSortOptions`. Pagination uses `SearchAfterValues([]types.FieldValue)` with a small `parseSearchAfter` helper that appends raw `int64`/`float64`/`string` values (since `types.FieldValue = any`). The test wire assertion is updated to match the typed Sort builder's long form (`{"_score": {"order": "desc"}}`), and mock responses now include the product header so subtests pass when run in isolation.
- **Transport/config demos**:
  - `security/tls_with_ca.go`, `security/tls_configure_ca.go`
  - `logging/default.go`, `logging/custom.go` — the intentional bad-query path now uses `Search().Q("{FAIL")` to reproduce the esapi `WithQuery` behaviour.
  - `interceptor/cmd/{auth_provider,context_auth,custom_auth,custom_observability}/main.go` — context-auth threads per-request context through `Do(ctx)` so the interceptor can read it.
  - `instrumentation/expvar.go` — typed fluent builders.
[...]

PR #1446, opened 2026-04-21
docs(migration): document (*Client).ToTyped() conversion method

MattDevy:
## What

Documents the `(*Client).ToTyped()` method added in #1445.

Two doc changes:

- **`docs/reference/low-level-api/migration.md`** — adds a new subsection under "Partial migration" titled "Get a full typed client from an existing low-level client", describing when to use `ToTyped`, how to call it, the one-shot-and-cache guidance, and the shared-close warning.
- **`docs/reference/typed-api/index.md`** — adds one sentence to the "Relationship to the low-level client" section pointing readers who already hold a `*elasticsearch.Client` at `ToTyped`, with a link into the migration guide.

## Why

#1445 adds the method but the migration guide was the natural place to explain it, and that PR deliberately left docs alone per our "one logical change per PR" rule.

The migration guide already covers two migration paths (full constructor swap, per-endpoint `search.New(client)` style). `ToTyped` is a third path that specifically helps the common case where the `*Client` is constructed in a shared infrastructure package the caller doesn't own. Without the doc, that option is easy to miss.

The shared-close behavior is the one sharp edge of the method — both clients point at the same transport, so `Close` on either tears down the pool used by both. Calling that out explicitly prevents a class of "why did my other client stop working" bugs.

## Review notes

- Depends on #1445. This can merge as soon as that lands — the docs describe a method that only exists on that branch.
- No code or tests changed. Pure docs PR.

## Test plan

- [x] Markdown renders correctly locally (headings, code blocks, anchors consistent with neighboring doc style).
- [x] Internal links resolve: the new subsection uses an anchor `[_to_typed]` consistent with the surrounding `_partial_migration_*` pattern.

PR #1445, opened 2026-04-21
feat(client): add (*Client).ToTyped() conversion method

MattDevy:
## What

Adds `(*Client).ToTyped() *TypedClient` so callers holding a `*elasticsearch.Client` can obtain a `*TypedClient` without rebuilding the transport.

```go
c, _ := elasticsearch.New(...)
tc := c.ToTyped()
_, _ = tc.Info().Do(ctx)
```

The returned `TypedClient` shares the source `Client`'s:

- `Transport` (and therefore its connection pool)
- `compatibilityHeader`, `autoDrainBody`, `disableMetaHeader` settings

## Why

We're promoting `TypedClient` as the preferred API surface, but a lot of existing code holds a `*Client` from legacy `NewClient`/`New` paths or from shared infrastructure packages. Today there's no way to hand those callers a `TypedClient` without building a whole new transport, which duplicates configuration and creates a second connection pool. `ToTyped` closes that gap.

## Design notes

- **No binary-size cost when unused.** `typedapi` is already an unconditional import of the root package (via `NewTypedClient`/`NewTyped`), so adding a method that references it does not grow the binary for programs that only use `*Client`.
- **Fields are enumerated, not struct-copied.** `BaseClient` holds a `sync.RWMutex` and a `closeDone uint32`; copying the struct wholesale would trip `go vet`. The method builds a fresh `BaseClient` by listing the plain fields, matching what `New`/`NewTyped` already do.
- **Idempotent.** Repeated conversions are safe (the internal meta-header suffix is only appended once).
- **Documented as a one-shot call to be reused.** Each invocation allocates a new `typedapi.MethodAPI` tree and the returned client will re-run the product check on its first request, so the godoc tells callers to cache the returned `*TypedClient` rather than calling it per-request.
- **Backwards compatible.** Purely additive: one new exported method on an existing type.

## Out of scope

- **Documentation under `docs/reference/`** will land in a follow-up PR per our "one logical change per PR" rule.
- **No reverse `(*TypedClient).ToFunctional()`**: not requested, and we're deliberately pushing toward typed, not away from it.

## Test plan

- [x] `make lint` clean
- [x] `make test-unit` — 324 tests pass
- [x] New `TestToTyped` subtests cover:
  - the outgoing client-meta header is updated to record typed-API usage
  - `tc.Transport == c.Transport` (identity, not just equivalence)
  - `compatibilityHeader`, `autoDrainBody`, `disableMetaHeader` are carried over
[...]

PR #1452, opened 2026-04-21
docs(examples): add built-in OTel and typed-migration examples

MattDevy:
## Summary

Follow-up to #1451 (now merged), which explicitly called these out as out-of-scope follow-up work. Adds three focused examples that round out the migration / observability story:

- **`_examples/instrumentation/otel.go`** — demonstrates the built-in OpenTelemetry hook, `elasticsearch.NewOpenTelemetryInstrumentation(provider, captureSearchBody)`, passed to the constructor via `WithInstrumentation(...)`. Self-contained: an inline `httptest` fake server plus a stdout trace exporter, so `go run otel.go` prints spans without needing a live cluster. The README section cross-links to the existing manual-interceptor demo at `_examples/interceptor/cmd/custom_observability` and explains when to reach for which: the built-in hook for standard tracing, interceptors when you need to extend behaviour.
- **`_examples/totyped.go`** — demonstrates `(*Client).ToTyped()` (landed in #1445 / documented in #1446) for incremental migration. Keeps the functional client handling one call, then `typed := es.ToTyped()` takes over for Info / Index / Search (using `esdsl.NewMatchQuery`). Notes that the shared transport means closing the functional client also closes the pool backing the typed one, and `defer`s `Close` accordingly. Uses `//go:build ignore` to match the `customization.go` pattern so it doesn't collide with `main.go`/`configuration.go` in the root `_examples` module.
- **`_examples/typed_endpoint.go`** — the lightest-weight migration step. Two variants in one file:
  - Pass the existing functional `*elasticsearch.Client` straight into a single typed endpoint constructor (here `search.New`), without constructing a `*TypedClient` or calling `ToTyped()`. Every typed endpoint package exports a `New(elastictransport.Interface)` constructor; the functional client satisfies that interface via the embedded `BaseClient.Perform` method, so there's no adapter and no second client.
  - For fresh code that only ever needs a handful of typed endpoints, `elasticsearch.NewBase(opts ...Option)` returns a raw `*BaseClient` directly, skipping both the `esapi` tree on `*Client` and the typedapi `MethodAPI` tree on `*TypedClient`. Pair it with `search.New(base)` (or any other endpoint package's `New`) for the minimum possible allocation.

Also updates the two `README.md` entries to cover the new files: the root `_examples/README.md` gains a "Migrating to the Typed API" section covering both typed-migration files and mentions built-in OTel alongside `expvar` under Instrumentation; `_examples/instrumentation/README.md` gets a dedicated "OpenTelemetry (built-in)" section that contrasts with the interceptor-based OTel demo at `_examples/interceptor/cmd/custom_observability`.

## Why

The library exposes `NewOpenTelemetryInstrumentation` as the first-class way to get OTel tracing, but until this PR the only OTel example we shipped was the interceptor-based `custom_observability` demo — great for understanding the mechanism, but misleading as a starting point for users who just want traces. Similarly, `ToTyped()`, direct-to-endpoint `New(client)`, and `NewBase()` all exist specifically to make migration (or fresh adoption) cheap, and short runnable recipes make those paths more discoverable than docs-only mentions.

## Reviewer notes

- Both root-level example files use `//go:build ignore` to match the existing `customization.go` pattern — the root `_examples` module has multiple `package main` files that only build one at a time via `go build <file>.go`.
- `_examples/totyped.go` uses `typedInfo.Version.Int` because `types.ElasticsearchVersionInfo` tags the server version string field as `Int json:"number"` (not `Number`). Mildly surprising but correct against the current generated types.
- `_examples/instrumentation/go.{mod,sum}` gained direct deps on `go.opentelemetry.io/otel`, `.../sdk`, and `.../exporters/stdout/stdouttrace` (pulled in by `go mod tidy` after adding `otel.go`). The module's direct-dep set is now strictly smaller than before #1451 (OpenCensus and Elastic APM are gone, replaced by three OTel modules the example actually uses).
- The `_examples/instrumentation` module still has the pre-existing vet noise from `expvar.go` (unbuffered signal channel, redundant newline) — unchanged by this PR. `go vet otel.go` is clean.
- Rebased on current `main` after #1451 landed; commits are split by theme (OTel+ToTyped / single-endpoint / NewBase) so they can be reviewed in order and are independently revertible pre-squash.

## Test plan

- [x] `make -C _examples/instrumentation test` builds both files (`expvar.go`, `otel.go`).
- [x] `go build -o /dev/null _examples/totyped.go` passes.
- [x] `go build -o /dev/null _examples/typed_endpoint.go` passes.
- [x] `go run _examples/instrumentation/otel.go` emits spans with the expected attributes (`db.system=elasticsearch`, `db.operation` set per call, `db.statement` set on the search span).
- [x] `make lint` clean at the repo level.
- [ ] Reviewer smoke of `totyped.go` / `typed_endpoint.go` against a running cluster if desired (functional path unchanged from existing `main.go`).


prodsecmachine:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/clients-team/pr-checks/1965e7d2-07e8-4bd6-bc09-c70c5f1a6337?refs=60b06a64d990760b19e0be082e643d7656135918&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/clients-team/pr-checks/1965e7d2-07e8-4bd6-bc09-c70c5f1a6337/license?refs=60b06a64d990760b19e0be082e643d7656135918&source=prChecksComment) |

_examples/instrumentation/otel.go
```diff
@@ -0,0 +1,127 @@
...
+	srv := newFakeES()
+	defer srv.Close()
+
+	// 3. Create the client with the built-in OTel instrumentation.
+	//    captureSearchBody = true includes the search query in the span
+	//    attributes for the search-family endpoints.
+	es, err := elasticsearch.NewTyped(
+		elasticsearch.WithAddresses(srv.URL),
+		elasticsearch.WithInstrumentation(
+			elasticsearch.NewOpenTelemetryInstrumentation(tp, true),
```
Anaethelion:
```suggestion
    captureSearchBody := true
	es, err := elasticsearch.NewTyped(
		elasticsearch.WithAddresses(srv.URL),
		elasticsearch.WithInstrumentation(
			elasticsearch.NewOpenTelemetryInstrumentation(tp, captureSearchBody),
```

Anaethelion (commented):
LGTM, left a quick nit for readability

_examples/instrumentation/otel.go
MattDevy: Good call — hoisted the literal into a named `captureSearchBody := true` local so the intent reads at the call site. Dropped in 60b06a64d9.

PR #1586, opened 2026-09-06
fix(esutil): reject expired contexts in BulkIndexer.Add

### LINKED ISSUE #1468: esutil: TestBulkIndexer/Add()_Timeout is flaky on Windows (Add() ignores already-expired context on non-blocking path)
`TestBulkIndexer/Add()_Timeout` fails intermittently on `windows-latest` with `-race`:

```
bulk_indexer_internal_test.go:380: Expected timeout error, but none in: [<nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil>]
```

Seen in: https://github.com/elastic/go-elasticsearch/actions/runs/25379904590/job/74425042637 (on `release-please--branches--9.4`; the identical branch passed 2 hours earlier in run 25373035721).

**Test intent:** Call `Add()` 10 times with an already-expired context and assert at least one returns `context.DeadlineExceeded`.

**Why it fails:** `Add()` only consults `ctx.Done()` in the final blocking `select` (reached only when all worker queues are full, `bulk_indexer.go:408`). The earlier non-blocking channel sends (`bulk_indexer.go:388-404`) ignore the context. With `NumWorkers: 1` and default `QueueSizeMultiplier: 1` (queue capacity = 1), the worker goroutine (using a no-op mock transport) drains the single-item buffer between `Add()` calls, so the non-blocking send always succeeds and the blocking select is never reached.

**Fix:** `Add()` should check `ctx.Done()` at entry, before any channel operation. An already-expired context should always return an error immediately, regardless of queue state.

```go
// At the top of Add(), before any channel send:
select {
case <-ctx.Done():
    if bi.config.OnError != nil {
        bi.config.OnError(ctx, ctx.Err())
    }
    return ctx.Err()
default:
}
```

This matches the expected behavior: if you pass a cancelled/expired context to `Add()`, you get an error back unconditionally.

IamPritamAcharya:
## Summary

Fixes #1468.

`BulkIndexer.Add` attempted its non-blocking worker enqueue paths before checking the caller's context. As a result, an item could be accepted when its context had already expired if a worker queue had available capacity.

Check `ctx.Err()` immediately after updating `NumAdded`, before item preparation and worker dispatch. This ensures an already-cancelled or expired context is returned immediately while preserving the existing `NumAdded` accounting and `OnError` behavior.

## Testing

Replaced the timing-dependent timeout test with a deterministic regression test that:

* uses an already-expired context;
* provides a worker queue with known spare capacity;
* verifies `context.DeadlineExceeded` is returned;
* verifies `OnError` is called exactly once with the context error;
* verifies `NumAdded` continues to count the attempted call;
* verifies item preparation is skipped;
* verifies the item is not sent to Elasticsearch.

Validation:

* `go test ./esutil/...`
* `go test -race ./esutil/...`
* `make test-unit`
* `make lint`

All unit and lint checks pass with the repository-pinned Go 1.25.12 toolchain.

### DIFF
--- a/esutil/bulk_indexer_internal_test.go
+++ b/esutil/bulk_indexer_internal_test.go
@@ -61,6 +61,18 @@ func (t *mockTransport) RoundTrip(req *http.Request) (*http.Response, error) {
 	return t.RoundTripFunc(req)
 }
 
+type errReadSeeker struct {
+	err error
+}
+
+func (r errReadSeeker) Read(_ []byte) (int, error) {
+	return 0, r.err
+}
+
+func (r errReadSeeker) Seek(_ int64, _ int) (int64, error) {
+	return 0, r.err
+}
+
 //nolint:gocyclo
 func TestBulkIndexer(t *testing.T) {
 	t.Run("Basic", func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


esutil/bulk_indexer_internal_test.go
```diff
@@ -351,34 +363,63 @@ func TestBulkIndexer(t *testing.T) {
...
+		es, err := elasticsearch.NewClient(elasticsearch.Config{Transport: &mockTransport{
+			RoundTripFunc: func(*http.Request) (*http.Response, error) {
+				requestCalls.Add(1)
+				return &http.Response{
+					StatusCode: http.StatusOK,
+					Body:       io.NopCloser(strings.NewReader(`{"took":1,"errors":false,"items":[]}`)),
+					Header:     http.Header{"X-Elastic-Product": []string{"Elasticsearch"}},
+				}, nil
+			},
+		}})
```
l-trotta: thanks for the contribution! just a nit before approving: the linter CI is failing because of NewClient used in the tests, which is deprecated. could you change it to this?

```suggestion
		es, err := elasticsearch.New(elasticsearch.WithTransportOptions(elastictransport.WithTransport(&mockTransport{
			RoundTripFunc: func(*http.Request) (*http.Response, error) {
				requestCalls.Add(1)
				return &http.Response{
					StatusCode: http.StatusOK,
					Body:       io.NopCloser(strings.NewReader(`{"took":1,"errors":false,"items":[]}`)),
					Header:     http.Header{"X-Elastic-Product": []string{"Elasticsearch"}},
				}, nil
			},
		})))
```

IamPritamAcharya:
@l-trotta Thanks for catching that! I’ve replaced the deprecated `NewClient` call with the functional-options constructor as suggested.

l-trotta (approved):
thank you :D

